PR #57, opened 2021-05-17
Tidy up use of brand name

karlchillmaid:
There are a few places where our approach to when to capitalise Emergency Alerts and when not to looks a bit weird.

This PR fixes that by making it more explicit when we’re talking about the service name.

It includes changes to page titles and navigation, so it *might* break some smoke tests.

## The rules are

`Emergency Alerts`

* landing page title
* breadcrumb for landing page
* footer menu heading
* only if we’re explicitly talking about the service, for example: ‘the Emergency Alerts service’

`emergency alerts` or `emergency alert`

* everywhere else

`alerts` or `alert`

* only use after ‘emergency alerts’ to reduce repetition and improve readability

### DIFF
--- a/src/how-alerts-work.html
+++ b/src/how-alerts-work.html
@@ -63,7 +63,7 @@
         Emergency alerts are free. You do not need to sign up for them or download an app.
       </p>
       <p class="govuk-body">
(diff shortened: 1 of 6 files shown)


benthorner (approved):
Here's the smoke test PR if you want to 👀 that: https://github.com/alphagov/notifications-broadcasts-infra/pull/256

quis (approved):
Ran and reviewed locally 👍🏻

PR #58, opened 2021-05-17
Remove example SVG from tab order in IE / Edge

36degrees:
Internet Explorer and Microsoft Edge up to version 13 treat SVGs as interactive elements and include them in the `tabindex` unless they have `focusable="false"`.

Add `focusable="false"` to the example SVG so that it's not included when tabbing in these browsers – at the minute if you tab forwards from the 'What happens when you get an alert' link it gets focused, but does not have a visible focus state so the focus appears 'lost'.


quis (approved):
This is some very specific knowledge 👍🏻

PR #60, opened 2021-05-18
Update iPhone instructions

karlchillmaid:
This PR updates the iPhone instructions based on feedback received on Twitter:

> Can I make a suggestion for http://gov.uk/alerts/when-you-get-an-alert? The iPhone instructions say “swipe down on the Home Screen”, but that’ll open Spotlight search if you swipe in the middle. A more reliable instruction would be “swipe down from the top of the screen”. Nice pages, very clear!

Would be great if at least 1 iPhone user could check this.

### DIFF
--- a/src/when-you-get-an-alert.html
+++ b/src/when-you-get-an-alert.html
@@ -85,7 +85,10 @@
       {% endset %}
       {% set iosAdvice %}
         <p class="govuk-body">
-            Swipe down from the top of the screen to view your notifications. If you clear your notifications you’ll delete the alert.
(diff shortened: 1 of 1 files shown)


pixeltrix:
These are the official Apple docs on how to do it: https://support.apple.com/en-gb/guide/iphone/iph6534c01bc/ios

idavidmcdonald:
I think this is an improvement on what we have but Andrews link for apple docs may give some further food for thought

karlchillmaid:
@pixeltrix @idavidmcdonald 

How’s this?

pixeltrix (approved):

PR #63, opened 2021-05-20
Fix visual bug with example SVG in IE11

tombye:
In IE9-11, it looks like removing the height and width attributes from the SVG made it render much smaller than it should be.

ie11_before

This replaces those attributes and rewrites the mobile breakpoint so it matches the intended size while maintaining the same aspect ratio.

### IE11

ie11_after

### Chrome

|Desktop|Mobile|
|---|---|
|with_max_height_desktop|iphone_8_chrome_emulation|

### DIFF
--- a/src/assets/stylesheets/components/example.scss
+++ b/src/assets/stylesheets/components/example.scss
@@ -3,8 +3,11 @@
 
 .example {
   display: block;
-  max-width: 315px;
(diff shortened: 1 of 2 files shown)


quis (approved):
Oops. Thanks for the fix 👍🏻

PR #64, opened 2021-05-20
Add some IDE and OS files to .gitignore

quis:
Copied from the admin app. Will stop me having to keep manually deleting those pesky OS X `.DS_Store` files.


benthorner (approved):
Looks like we have these in a few other repo already 👍. Since the list of these is pretty much unbounded, we could also all consider having a global `~/.gitignore`, which can prevent this sort of thing in general, for each of our specific environments.

PR #65, opened 2021-05-20
Add current alert for Sizewell demo

benthorner:

Note that all the content and imagery in this PR is provisional.

TODO:

- [x] Add provisional content for headline / description
- [x] Confirm map area to use (currently East Suffolk)
- [x] Confirm approach for alert page URL
- [x] Figure out approach for showing alert after expiry

## Screenshots

| Alert list | Alert page |
| Screenshot 2021-05-24 at 13 49 15 | Screenshot 2021-05-24 at 13 49 32 |


idavidmcdonald:
image
image

Just spun this up locally and going to point out a few potential things that I'd like opinions on from @quis and @karlchillmaid 

1. Does `Sizewell` mean anything to users. It is a name of a tiny hamlet. Is that confusing without something like 'Sent to Sizewell' or 'Sent to the area of Sizewell' or 'Sent to Sizewell in East Suffolk'

2. Is the styling on the word Sizewell on the current alerts page what we want? Looks a little strange how it is much smaller than the content and made to look a little stranger given it is a single word

3. Do we want to force 'life-threatening' to be on a single line rather than split across two lines. See non breaking hyphen character mentioned in https://stackoverflow.com/questions/8753296/how-to-prevent-line-break-at-hyphens-on-all-browsers

data.yaml
```diff
@@ -1,2 +1,16 @@
-last_updated: 2021-04-27T13:30:01-00:00
-alerts: []
+last_updated: 2021-05-25T13:00:00+01:00
+alerts:
+  - identifier: sizewell-demo
+    message_type: alert
+    sent: 2021-05-25T13:00:00+01:00
+    expires: 2021-05-25T14:00:00+01:00
+    headline: Test Emergency Alert
```
idavidmcdonald, line 7: @karlchillmaid one here to check in terms capitalisation

src/index.html
```diff
@@ -42,9 +42,6 @@ <h1 class="govuk-heading-xl">
   {% if current_alerts %}
   <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
   {{ banner(current_alerts | length, data_last_updated) }}
-  {% else %}
-    <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
-    {{ planned_tests_banner(1, "Tuesday 25 May") }}
```
idavidmcdonald, line 47: Good spot Ben, this is good that we've removed it as after testing this link should get removed.

@quis and @karlchillmaid not that important but could be worth thinking what should be shown on the planned tests page (even though we won't link to it directly) during and after the current alert

idavidmcdonald (commented):
So @benthorner this looks great thank you.

I don't see anything that needs a technical change here or anything that you can decide on your own. But I have left a few comments for Chris and Karl for some design/content thoughts.

I won't leave an approving review at this point because I suspect they may ask for a few changes and this would need to be reviewed again

idavidmcdonald:
Oh and another one from me, this pull request doesn't cover what happens when the alert expires. Wasn't there supposed to be some page of previous alerts that this appears on? Didn't we use to have links in the footer to 'current alerts' and 'previous alerts'? I can't seem to find a previous alerts page in our code history for this so maybe it was just on the prototype. Probably a question for @quis

benthorner:
Thanks for checking it over @idavidmcdonald!

> this pull request doesn't cover what happens when the alert expires

Yes, that's deliberate. This morning we agreed the plan is to:

- Build "previous alerts" as a separate feature
- Let it auto-activate when the alert expires (except we will need to manually trigger a rebuild)

@quis has just approved [an initial PR I raised to making adding the feature easier](https://github.com/alphagov/notifications-govuk-alerts/pull/70). Once that's in, I'll rebase this one and then go back to adding the feature itself.

data.yaml
```diff
@@ -1,2 +1,16 @@
...
+    description: |
+      The UK government is testing Emergency Alerts in East Suffolk.
+
+      Emergency alerts tell you what to do if there’s a life-threatening event nearby.
+
+      To find out more, call 0808 1697692 or search for gov.uk/alerts
+    static_map_png: map-sizewell-demo-static.png
+    area_names:
+      - Sizewell
```
quis, line 16: For consistency with how the test is being talked about in publicity, and on our planned alerts page:
```suggestion
      - East Suffolk
```

src/index.html
quis: PR prepared for after East Suffolk here: https://github.com/alphagov/notifications-govuk-alerts/pull/74

data.yaml
benthorner: Sure, I've amended that in the first commit.

src/index.html
benthorner: I've dropped the commit to remove the else clause. If we deploy #74 between 1-2PM, then it'll show up automatically when the alert expires.

data.yaml
```diff
@@ -1,2 +1,17 @@
...
+    headline: Test Emergency Alert
+    description: |
+      The UK government is testing Emergency Alerts in East Suffolk.
+
+      Emergency alerts tell you what to do if there’s a life-threatening event nearby.
+
+      To find out more, call 0808 1697692 or search for gov.uk/alerts
+    static_map_png: map-sizewell-demo-static.png
+    area_names:
+      - East Suffolk
```
quis, line 17: For consistency with the new map image, and to avoid people elsewhere in East Suffolk wondering why they haven’t got the alert:
```suggestion
      - Leiston in East Suffolk
```

benthorner: Updated in the first commit 👍.

idavidmcdonald (approved):

PR #67, opened 2021-05-20
Add survey component

tombye:
Adds a survey component, loosely based on [GOVUK's feedback component](https://components.publishing.service.gov.uk/component-guide/feedback).

https://www.pivotaltracker.com/story/show/177726051

|Desktop|Mobile|
|---|---|
|survey_desktop|survey_mobile|

Currently contains fake content so I'm leaving this in draft until @karlchillmaid can write something real for it.

## Notes for reviewers

I've checked on all browsers in our matrix and the link in it has been checked against the [GOVUK feedback component's accessibility criteria](https://components.publishing.service.gov.uk/component-guide/feedback).

GOVUK's sits outside the `<main>` element. That would require some slightly hacky workarounds for us because the GOVUK Frontend template, which we use for our layout, doesn't have a slot for that point in the HTML. Because of this, I've included it in the `<main>` element which, I think, has relatively low impact on the structure of the document.

It includes a hidden `<h2>`, which was a suggestion by the GOVUK accessibility team to make it easier to find in the document structure for non-visual users.

This PR also adds a blue border to the top of the footer to match that on the rest of www.gov.uk.

### DIFF
--- a/src/404.html
+++ b/src/404.html
@@ -2,7 +2,7 @@
 
 {% block pageTitleCurrent %}Content not found{% endblock %}
 
-{% block content %}
+{% block mainContent %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-xl">
--- a/src/alert.html
+++ b/src/alert.html
@@ -44,7 +44,7 @@
   }) }}
 {% endblock %}
 
-{% block content %}
+{% block mainContent %}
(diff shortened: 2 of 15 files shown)


benthorner (approved):
Looks good from a technical perspective, thanks for explanations 📚. 

I'll approve it now, and we can tweak the survey link when we're ready.

quis:
Desktop | Mobile
---|---
image | image

PR #68, opened 2021-05-20
Update content for ‘How to opt out of emergency alerts’

karlchillmaid:
This PR updates the introduction to the opt out page.

We’re making this change because it’s no longer true to say that users can opt out of all alerts.

## Background
The 25 May test will be sent on the ‘severe’ and ‘government’ channels – the latter of which is mandatory.

Currently we do not mention the ‘government’ channel because:

- explaining the different channels makes the content more complex
- the policy decision about if and when ‘government’ alert will not happen until after the East Suffolk test

Until that decision is made we do not want to introduce an explanation of the channels. However, we should update the content so it’s closer to the statement given to the press, which said:

> “It will be possible to opt-out of some alerts through the phone's settings. However, the most important alerts will always come through and the government recommends that people do not opt-out.”

### DIFF
--- a/src/opt-out.html
+++ b/src/opt-out.html
@@ -39,10 +39,10 @@
         How to opt out of emergency alerts
       </h1>
       <p class="govuk-body">
-        If you opt out, you will not get any emergency alerts.
+        You can opt out of some emergency alerts, but you’ll always get the most important ones.
       </p>
       <p class="govuk-body">
-        You cannot choose the type of hazard you want to get alerts about. If you opt out because you do not want flood warnings for example, you’ll also miss alerts for fires and terrorist incidents.
+        You cannot opt out by subject, only by how serious the emergency is. If you opt out because you do not want flood warnings, for example, you’ll also miss alerts for fires and terrorism.
       </p>
       <p class="govuk-body">
         Because of this, you should keep emergency alerts switched on for your own safety.
--- a/src/opt-out.html
+++ b/src/opt-out.html
@@ -39,7 +39,7 @@
         How to opt out of emergency alerts
       </h1>
       <p class="govuk-body">
-        You can opt out of some emergency alerts, but you’ll always get the most important ones.
+        You can opt out of some emergency alerts, but not the most important ones.
       </p>
       <p class="govuk-body">
         You cannot opt out by subject, only by how serious the emergency is. If you opt out because you do not want flood warnings, for example, you’ll also miss alerts for fires and terrorism.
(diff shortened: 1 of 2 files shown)


src/opt-out.html
```diff
@@ -39,10 +39,10 @@ <h1 class="govuk-heading-xl">
         How to opt out of emergency alerts
       </h1>
       <p class="govuk-body">
-        If you opt out, you will not get any emergency alerts.
+        You can opt out of some emergency alerts, but you’ll always get the most important ones.
```
karlchillmaid, line 42: **Alternative**

Which do we think is clearer?

```suggestion
        You can opt out of all but the most important emergency alerts.
```

karlchillmaid:
```suggestion
        You can opt out of some emergency alerts, but not the most important ones.
```

src/opt-out.html
```diff
@@ -39,10 +39,10 @@ <h1 class="govuk-heading-xl">
         How to opt out of emergency alerts
       </h1>
       <p class="govuk-body">
-        If you opt out, you will not get any emergency alerts.
+        You can opt out of some emergency alerts, but not the most important ones.
       </p>
       <p class="govuk-body">
-        You cannot choose the type of hazard you want to get alerts about. If you opt out because you do not want flood warnings for example, you’ll also miss alerts for fires and terrorist incidents.
+        You cannot opt out by subject, only by how serious the emergency is. If you opt out because you do not want flood warnings, for example, you’ll also miss alerts for fires and terrorism.
```
karlchillmaid, line 45:
```suggestion
        You cannot opt out by subject, only by how serious the emergency is. If you opt out because you do not want flood warnings, for example, you might miss alerts for fires and terrorism.
```

idavidmcdonald (approved):
So content wise this is all good.

I am a bit cautious about leaving this content up on there until a decision is made about whether the government channel will be used. That could not be until July. Don't want people to get worried about not being able to opt out when it maybe gets decided that we won't use the government channel.

I suggest a second opinion, maybe @quis if he has time...

karlchillmaid:
@idavidmcdonald

 I think you’re right to be cautious. I’m also worried about the opposite:
 
Leaving the content as is, telling people they can opt out of all alerts, then finding out that the government channel will be used.

Also, a decision _not_ to use the government channel is not irreversible. It’s possible that a scenario exists that would cause that decision to be changed at some point in the future.

idavidmcdonald:
Your point about the opposite has swayed me. I think this is the lesser of two evils 👍 

My approval stands. Feel free to get another opinion if you want one but from my point, this is good!

quis (approved):

PR #69, opened 2021-05-20
Test that dates in alert data have correct time offsets

CrystalPea:
So that they respect daylight saving time changes.

This implements following decision:

Make all timestamps iso 8601, local to the Europe/London with an explicit offset*.

For example: 10:00am on 12 March 2021, which is outside of British Summer Time (BST)
would be:
2021-03-12T10:00+00:00

Whereas 10:00am on 12 May 2021, which is inside BST would be:
2021-05-12T10:00+01:00

This would be enforced by:

a test that checks all the timestamps in data.yaml
an example alert which explains the format used and variations to consider
*following the formatting rules in
https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators

Pivotal ticket: https://www.pivotaltracker.com/story/show/178198145

### DIFF
--- a/tests/test_data_yaml.py
+++ b/tests/test_data_yaml.py
@@ -1,4 +1,7 @@
+from datetime import datetime
+
 import pytest
+import pytz
 
 from lib.alerts import Alerts
 from lib.utils import REPO
@@ -13,3 +16,18 @@ def test_alert_sent_before_starts_before_expires(alerts):
     for alert in alerts:
         assert alert.sent_date.as_utc_datetime <= alert.starts_date.as_utc_datetime
         assert alert.starts_date.as_utc_datetime < alert.expires_date.as_utc_datetime
+
+
+def is_date_in_london_timezone_including_summertime(date):
+    if not isinstance(date, datetime):
+        return False
+    zoneless_date = date.replace(tzinfo=None)
+    date_with_expected_offset = pytz.timezone('Europe/London').localize(zoneless_date)
+    return date_with_expected_offset.utcoffset() == date.utcoffset()
+
+
+def test_dates_in_alerts_data_include_explicit_timezone_offset(alerts):
+    for alert in alerts:
+        assert is_date_in_london_timezone_including_summertime(alert.sent)
+        assert is_date_in_london_timezone_including_summertime(alert.starts)
+        assert is_date_in_london_timezone_including_summertime(alert.expires)


tests/test_datayaml.py
```diff
@@ -0,0 +1,35 @@
...
+def get_alerts_data(path):
+    with open(path) as file:
+        alerts_data = yaml.load(file, Loader=yaml.FullLoader)
+    return alerts_data
+
+
+def does_date_contain_explicit_offset(date):
+    if not isinstance(date, datetime):
+        return False
+    localised_datetime = pytz.timezone('Europe/London').localize(date.replace(tzinfo=None))
```
benthorner, line 20: Minor: we could make this a bit clearer / more explicit:

```
zoneless_date = date.replace(tzinfo=None)
expected_offset = zoneless_date.astimezone(timezone('Europe/London')).utcoffset()
actual_offset = date.utcoffset()
```

tests/test_datayaml.py
```diff
@@ -0,0 +1,35 @@
...
+    return localised_datetime.utcoffset() == date.utcoffset()
+
+
+@pytest.mark.parametrize('yaml_file_path', [
+    'tests/test_files/example_alert_data.yaml',  # example alert that shows how dates should be formatted
```
benthorner, line 25: We don't need a separate example file. The "example" mentioned in the card was meant to be a comment in `data.yaml`, with guidance to change the offset when copy/pasting. The assertion here is even better, because we actually check for the expected offset. Soon, we will also have real alerts to copy from.

It might be easier to split this in two:

- One test for the actual data file.

- One test function for the assertion code, which could mock the YAML load bit to test a few different pieces of data. Or we could just trust that the assertion code works. If we do want to test the test 😄, then we should have some test data that we expect to fail as well.

tests/test_files/example_alert_data.yaml
```diff
@@ -0,0 +1,16 @@
...
...20760012, -7.124037548551484], [54.09121885239279, -7.308294258459859], [54.10783178248695, -7.583960428694487], [54.18539905806525, -7.845223966277847], [54.44579647131303, -8.192958519674985]], [[57.75245668793269, -8.592820720527689], [57.798349418800925, -8.691519032735506], [57.86455485323485, -8.684557268975732], [57.91418787823128, -8.457317784625603], [57.83056845766009, -8.444037273308467], [57.75245668793269, -8.592820720527689]], [[59.112525647448926, -5.882765486506044], [59.181542047045404, -5.81291199632521], [59.10774665753379, -5.764033432697249], [59.065961116966065, -5.850962305785022], [59.112525647448926, -5.882765486506044]], [[60.12608338654564, -2.165529386790981], [60.34563034632679, -1.8219420500179084], [60.66077341109561, -1.470478899909446], [60.88793434492865, -0.9255934119305854], [60.84093307381703, -0.7139984990943649], [60.41704323871041, -0.6902305720104588], [60.27870185536764, -0.7717506307935432], [59.47160239287521, -1.5998545261997428], [59.508860947732025, -1.7073761021153309], [59.90779481600101, -1.6824886013593097], [60.09681539380499, -1.9172217831085037], [60.05915899413415, -2.0650519214591108], [60.12608338654564, -2.165529386790981]]]
```
benthorner, line 7: This isn't accurate anymore.

Here's the YAML we're planning to use for the Sizewell demo: https://github.com/alphagov/notifications-govuk-alerts/pull/65/files#diff-3d941c62395536004b33c30eb9f00d6b763d619dad61fb7522b8a3299e88a981

tests/test_datayaml.py
```diff
@@ -0,0 +1,35 @@
...
+
+from datetime import datetime
+
+
+KEYS_CONTAINING_DATES = ['last_updated', 'expires', 'sent']
+
+
+def get_alerts_data(path):
+    with open(path) as file:
+        alerts_data = yaml.load(file, Loader=yaml.FullLoader)
```
benthorner, line 13: Minor: we can just do the following, which would make it easier to mock the loading (related to my comment about the test 👇).

```
return yaml.load(file, Loader=yaml.FullLoader)
```

The `with` statement is only really necessary for long running processes, and even then, Python will usually clean up the file handle at some point.

tests/test_datayaml.py
```diff
@@ -0,0 +1,35 @@
...
+    return alerts_data
+
+
+def does_date_contain_explicit_offset(date):
```
benthorner, line 17: Minor: this could be more specific e.g. `is_date_in_london_timezone`. Similarly for the name of the test function.

benthorner (commented):
Nice work with the assertion 💯. That's much better hoping people will change the offset when they copy/paste an example/previous alert. I think there's been a bit of miscommunication about what "example" meant in the card, so we can simplify the tests a bit and avoid having to keep up with the format of the `data.yaml` file.

tests/test_datayaml.py
CrystalPea: ok, will do!

tests/test_datayaml.py
CrystalPea: ok,  I will get rid of test data, I don't think we need to test the test 😅

tests/test_files/example_alert_data.yaml
CrystalPea: useful, saw one additional key (`"starts"`) there

tests/test_datayaml.py
CrystalPea: done ^_^

benthorner (approved):
Looks good 👍.

Just a linting error to fix.

PR #70, opened 2021-05-21
Refactor data handling to support previous alerts

benthorner:

In attempting to add a previous alerts feature I got stuck
because the "is_alert_current" function could only be used
for a limited time, before any rendering took place.

This PR makes a broad refactoring to cut down on the number
of utilities, so all the data and functionality is easier to
find, and always works together.

### DIFF
--- a/lib/alert.py
+++ b/lib/alert.py
@@ -1,0 +1,10 @@
+from lib.alert_date import AlertDate
+
+
+class Alert:
+    def __init__(self, dict_):
+        for key, value in dict_.items():
+            setattr(self, key, value)
+
+        self.sent = AlertDate(self.sent)
+        self.expires = AlertDate(self.expires)
--- a/lib/alert.py
+++ b/lib/alert.py
@@ -1,4 +1,6 @@
 from lib.alert_date import AlertDate
+import pytz
+from datetime import datetime
 
 
 class Alert:
@@ -8,3 +10,11 @@ class Alert:
 
         self.sent = AlertDate(self.sent)
         self.expires = AlertDate(self.expires)
+
+    @property
+    def is_current(self):
+        now = datetime.now(pytz.utc)
+
+        if self.expires.as_utc_datetime <= now:
+            return False
+        return True
--- a/lib/alert.py
+++ b/lib/alert.py
@@ -1,22 +1,34 @@
 from datetime import datetime
 
 import pytz
+from notifications_utils.serialised_model import SerialisedModel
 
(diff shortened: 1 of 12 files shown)


lib/alert.py
```diff
@@ -0,0 +1,22 @@
...
+from lib.alert_date import AlertDate
+
+
+class Alert:
+    def __init__(self, dict_):
+        for key, value in dict_.items():
+            setattr(self, key, value)
+
+        self.sent = AlertDate(self.sent)
+        self.expires = AlertDate(self.expires)
```
quis, line 14: You could use `SerialisedModel` from utils to keep this consistent with our other repos:

```suggestion
class Alert(SerialisedModel):
    ALLOWED_PROPERTIES = {
        'identifier',
        'message_type',
        'sent',
        'expires',
        'headline',
        'description',
        'static_map_png',
        'area_names',
    }

    @property
    def sent_date(self):
        return AlertDate(self.sent)

    @property
    def sent_date(self):    
        return AlertDate(self.expires)
```

lib/alert.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @property
+    def is_current(self):
+        now = datetime.now(pytz.utc)
+
+        if self.expires.as_utc_datetime <= now:
+            return False
+        return True
```
quis, line 22: And you could use `SerialisedModelCollection` to avoid manually appending to the list in `build.py`

```suggestion
        return True
        
        
class Alerts(SerialisedModelCollection):
    model = Alert        
```

Example usage:
```python
alerts = Alerts(data['alerts'])
```

lib/alert.py
quis: The main advantage of this is that it guards against accidentally leaving out properties in the YAML. Jinja is very happy to fail silently if a property is missing.

benthorner: Oh nice! I didn't realise we had the utils repo already. Having a separate property name for the converted dates also helps prevent confusion between the two types.

I've added a commit to convert to what you suggested 👍.

quis (approved):
Good refactoring 💪🏻

lib/alert.py
benthorner: I was in two minds about this, but then realised a few more benefits:

- We can use it to do the YAML loading
- We can use it to encapsulate the `last_updated_at` conversion
- We can use it to encapsulate filtering current alerts

I've made your suggested change and the `last_updated_at` one in a new commit, and added a further one to extract the current alert filter.

PR #71, opened 2021-05-21
Lint frontend code as part of test, not build step

quis:
This matches where we check Python code with `flake8`. It means you don’t have to spend time waiting for the lint task to run every time you want to rebuild the site.


benthorner (approved):
Makes sense 👍

PR #72, opened 2021-05-21
Add a minimal page for previous alerts

benthorner:

I've tried to limit the number of decisions in this PR. Once
this is merged, it should then be a safe and simpler task to
iterate the design, noting that the page won't have content
until the first alert expires.

Things to (ideally) do in future work:

- Rebase https://github.com/alphagov/notifications-govuk-alerts/pull/65 to support "paragraphize" on the previous alerts page.
- Decide if we want to show author/timing info in the list of previous/current alerts
- Decide if we want to modify the single alert page to clearly show it's expired (we probably want to do more than change the colour of the icon, so I've just left it untouched in this PR)

## Screenshots (pages)

| New page  | Expired alert page (breadcrumb) |
| Screenshot 2021-05-21 at 17 50 31 | Screenshot 2021-05-24 at 10 34 22
 |

## Screenshot (footer)

Screenshot 2021-05-24 at 10 35 21

### DIFF
--- a/src/alert.html
+++ b/src/alert.html
@@ -21,6 +21,12 @@
   {{ pageTitle }}
 {%- endblock %}
 
+{% if alert_data.is_expired %}
+  {% set parentBreadcrumb = {"text": "Past alerts", "href": "/alerts/past-alerts"} %}
+{% else %}
+  {% set parentBreadcrumb = {"text": "Current alerts", "href": "/alerts/current-alerts"} %}
+{% endif %}
+
 {% block beforeContent %}
   {{ govukBreadcrumbs({
     "items": [
@@ -32,17 +38,14 @@
         "text": "Emergency alerts",
         "href": "/alerts"
       },
-      {
-        "text": "Current alerts",
-        "href": "/alerts/current-alerts"
-      }
+      parentBreadcrumb,
     ]
   }) }}
 {% endblock %}
 
 {% block content %}
   <h1 class="govuk-heading-xl alerts-icon__container alerts-icon__container--48">
-    {{ alerts_icon(height=48) }}
+    {{ alerts_icon(height=48, alert_active=alert_data.is_current) }}
     {{ alert_data.headline }}
   </h1>
   <div class="govuk-grid-row">
--- a/templates/_layout.html
+++ b/templates/_layout.html
@@ -47,6 +47,14 @@
         'columns': 2,
         'items': [
           {
+            'text': 'Current alerts',
+            'href': '/alerts/current-alerts'
+          },
+          {
+            'text': 'Past alerts',
+            'href': '/alerts/past-alerts'
+          },
+          {
             'text': 'When you get an emergency alert',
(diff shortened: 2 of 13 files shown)


quis:
Can we rename this everywhere (page titles, URLs, etc) to ‘Past alerts’? This is what Karl renamed the page in the prototype as, see https://alerts.cloudapps.digital/previous

src/alert.html
```diff
@@ -21,6 +21,12 @@
   {{ pageTitle }}
 {%- endblock %}
 
+{% if alert_data.is_expired %}
```
quis, line 24: Can we also set the icon to grey on this page is the alert is expired?

templates/_layout.html
```diff
@@ -61,6 +61,10 @@
           {
             'text': 'Opt out of emergency alerts',
             'href': '/alerts/opt-out'
+          },
+          {
+            'text': 'Previous alerts',
+            'href': '/alerts/previous-alerts'
           }
```
quis, line 68: We should include the link to current alerts because the two are complementary, even though the page will be empty for now:
```suggestion
          },
          {
            'text': 'Current alerts',
            'href': '/alerts/current-alerts'
          },
          {
            'text': 'Previous alerts',
            'href': '/alerts/previous-alerts'
          }
```

src/alert.html
benthorner: Sure, I've amended the last commit to do that 👍.

templates/_layout.html
benthorner: Updated - I've put them as the first two links, as in the prototype.

benthorner:
Thanks for taking a look @quis. I've amended the last commit to rename the page title and the URL, and amended the screenshots to match. Do you think we can defer a content review and get it in this morning?

templates/_layout.html
```diff
@@ -46,6 +46,14 @@
...
         'columns': 2,
         'items': [
+          {
+            'text': 'Past alerts',
+            'href': '/alerts/past-alerts'
+          },
+          {
+            'text': 'Current alerts',
+            'href': '/alerts/current-alerts'
+          },
```
quis, line 56: Current should come before past, to match the prototype (and because stuff on the web is typically ordered reverse chronologically).

Other than this I think the PR is good to go. I’ve spun it up locally and had a good click around.

```suggestion
          {
            'text': 'Current alerts',
            'href': '/alerts/current-alerts'
          },
          {
            'text': 'Past alerts',
            'href': '/alerts/past-alerts'
          },
```

quis (approved):

templates/_layout.html
benthorner: OK. I've amended the commit to reverse the ordering.

benthorner:
@quis thanks for approving. What would you like to do about these two follow-up tasks?

- Decide if we want to show author/timing info in the list of previous/current alerts
- Decide if we want to modify the single alert page to clearly show it's expired (we probably want to do more than change the colour of the icon)

PR #74, opened 2021-05-24
Add details of planned test in Reading

quis:
_Should not be merged until after 1pm on Tuesday 25 May._

***

So that this page doesn’t become outdated this commit replaces details of the East Suffolk test (which will be on the previous alerts page) with details of the upcoming test in Reading.

The date of this test has been given to the media<sup>1</sup>, so it’s OK for us to give the date publicly. We can add more detail closer to the day.

---

1. https://www.bbc.co.uk/news/uk-politics-57145675

### DIFF
--- a/src/planned-tests.html
+++ b/src/planned-tests.html
@@ -39,31 +39,17 @@
         {{ pageTitle }}
       </h1>
       <h2 class="govuk-heading-m govuk-!-margin-bottom-2">
-        Tuesday 25 May 2021
+        Tuesday 15 June 2021
       </h2>
       <h3 class="govuk-body govuk-!-margin-bottom-6">
-        East Suffolk
+        Reading
       </h3>
       <p class="govuk-body">
-        The UK government will send 2 test alerts between 1pm and 2pm.
+        The UK government will send a test alert.
       </p>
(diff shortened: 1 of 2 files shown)


src/planned-tests.html
```diff
@@ -39,34 +39,17 @@ <h1 class="govuk-heading-xl">
         {{ pageTitle }}
       </h1>
       <h2 class="govuk-heading-m govuk-!-margin-bottom-2">
-        Tuesday 25 May 2021
+        Tuesday 15 June 2021
```
benthorner, line 42: Do we need to change this [here](https://github.com/alphagov/notifications-govuk-alerts/blob/3eca64d72e36d704592ce2a614a2a2c2eeb4b2ff/src/index.html#L47) as well?

Optional: to make our lives easier, we could set the date in code and use it to say when the banner and page content should disappear. We could even set it as a field in the YAML file, so it's a global variable we don't forget about.

quis: We’re removing the homepage banner once the East Suffolk alert goes out. We’ll probably bring it back a week before Reading. Would support making it more automated at that point.

benthorner (approved):

PR #76, opened 2021-05-24
Use 'sent' property to show alerts early

benthorner:

## Screenshots (showing future timestamps)

| Alert page  | Home page |
| Screenshot 2021-05-24 at 10 51 16 | Screenshot 2021-05-24 at 10 51 21 |

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -12,6 +12,7 @@ def alert_dict():
         'area_names': [],
         'message_type': 'alert',
         'identifier': '1234',
+        'starts': datetime(2021, 4, 21, 11, 25, tzinfo=pytz.utc),
         'sent': datetime(2021, 4, 21, 11, 30, tzinfo=pytz.utc),
         'expires': datetime(2021, 4, 21, 12, 30, tzinfo=pytz.utc)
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -12,6 +12,7 @@ def alert_dict():
         'area_names': [],
         'message_type': 'alert',
         'identifier': '1234',
+        'starts': datetime(2021, 4, 21, 11, 25, tzinfo=pytz.utc),
```
rjbaker, line 15: This is a bit picky, but I think we've swapped the meaning of `starts/onset` and `sent` dates when compared to their use in CAP which might get a little confusing.

In CAP the sent date is when the alert is displayed, the onset/starts date is when the thing the alert pertains to begins. In this case I think the "sent" date would be used to determine visibility of the alert.

benthorner: Nice catch @rjbaker! I think it's important we're consistent with terminology, so we don't make mistakes. I've added a third commit to swap the meaning of the two fields, with a test to check they behave as expected.

rjbaker (approved):

PR #82, opened 2021-05-24
Set the last updated automatically

quis:
The most recent time the site will be updated with new information is when the most recent alert was broadcast.

This commit sets that date automatically, so we don’t have to update it manually (something we might forget to do).

### DIFF
--- a/lib/alerts.py
+++ b/lib/alerts.py
@@ -8,10 +8,6 @@ from lib.alert_date import AlertDate
 class Alerts(SerialisedModelCollection):
     model = Alert
 
-    def __init__(self, data):
-        self.last_updated = data['last_updated']
-        super().__init__(data['alerts'])
-
     @property
     def current(self):
         return [alert for alert in self if alert.is_current]
@@ -19,6 +15,10 @@ class Alerts(SerialisedModelCollection):
     @property
     def expired(self):
         return [alert for alert in self if alert.is_expired]
+
+    @property
+    def last_updated(self):
+        return max(alert.starts for alert in self if alert.is_current)
 
     @property
     def last_updated_date(self):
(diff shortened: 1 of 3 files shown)


tests/test_alerts.py
```diff
@@ -25,6 +27,30 @@ def test_from_yaml_loads_data(tmp_path, alert_dict):
     assert isinstance(alerts.last_updated_date, AlertDate)
 
 
+@freeze_time(datetime(
+    2021, 4, 21, 11, 30, tzinfo=pytz.utc
+))
+def test_last_updated(alert_dict):
```
benthorner, line 33: Minor: could we test as much with just two alerts?

lib/alerts.py
```diff
@@ -20,6 +16,10 @@ def current(self):
     def expired(self):
         return [alert for alert in self if alert.is_expired]
 
+    @property
+    def last_updated(self):
+        return max(alert.starts for alert in self if alert.is_current)
```
benthorner, line 21: Minor: we could test to show that we expect to get an exception here in the case there are no alerts (since this code shouldn't be used in that scenario).

benthorner (approved):
Nice 🇨🇫. Just a couple of minor / optional comments...

idavidmcdonald (approved):
Looks good I think.

Just something to check you are aware of

Tomorrow, for between roughly 12:55 and 13:00 it will say 'Updated 25 May 2021 at 13:00'. This will essentially say updated in the future.

Also another thing to check, is 'Updated 25 May 2021 at 13:00' the correct format for date times to be shown to users? @karlchillmaid ? I see https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#times which isn't explicitly date times but does imply we might need to change the formatting? Note, this probably wouldn't stop your PR going in as can be followed up with a later one on how to format the datetime

lib/alerts.py
quis: Added test in f04c6ff

tests/test_alerts.py
quis: Changed in 310dbca

PR #83, opened 2021-05-24
Update what happens when you get an emergency alert

karlchillmaid:
This PR updates the list of things that can happen when you get an emergency alert.

In previous trials, a higher proportion of devices than expected read the alert aloud. We should make users aware of this.

We’ve also changed ‘will’ to ‘may’, because not all devices behave the same way when they get an alert.

### DIFF
--- a/src/when-you-get-an-alert.html
+++ b/src/when-you-get-an-alert.html
@@ -42,7 +42,7 @@
       <p class="govuk-body">
         Your phone or tablet may: 
       </p>
-      <ul class="govuk-list govuk-bullet-list">
+      <ul class="govuk-list govuk-list--bullet">
         <li>
            make a loud siren-like sound, even if it’s set on silent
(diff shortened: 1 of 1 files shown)


src/when-you-get-an-alert.html
```diff
@@ -40,10 +40,21 @@ <h1 class="govuk-heading-xl">
         When you get an emergency alert
       </h1>
       <p class="govuk-body">
-        Your phone or tablet will vibrate and make a loud siren-like sound, even if it’s set on silent.
+        Your phone or tablet may: 
       </p>
+      <ul class="govuk-list govuk-bullet-list">
```
karlchillmaid, line 45:
```suggestion
      <ul class="govuk-list govuk-list--bullet">
```

benthorner (approved):

PR #88, opened 2021-05-25
Format dates and times in GOV.UK style

quis:
This pull request makes the way we show dates and times conform to the GOV.UK styleguide<sup>1</sup>. It also makes the date on the page for a single alert absolute, not relative.

# Before 

image

# After 

image

***

1. https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#times

### DIFF
--- a/lib/alert_date.py
+++ b/lib/alert_date.py
@@ -10,10 +10,19 @@ class AlertDate(object):
         self._local_datetime = _datetime.astimezone(timezone('Europe/London'))
 
     @property
+    def time_as_lang(self):
+        return {
+            '12:00AM': 'Midnight',
+            '12:00PM': 'Midday'
+        }.get(
+            f'{self._local_datetime:%-I:%M%p}',
+            f'{self._local_datetime:%-I:%M%p}',
+        ).lower()
+
+    @property
     def as_lang(self, lang='en-GB'):
         dt = self._local_datetime
-        am_pm = f'{dt:%p}'.lower()
-        return f'at {dt:%-I}:{dt:%M}{am_pm} on {dt.day} {dt:%B} {dt:%Y}'
+        return f'at {self.time_as_lang} on {dt.day} {dt:%B} {dt:%Y}'
 
     @property
     def as_iso8601(self):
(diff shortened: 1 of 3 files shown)


lib/alert_date.py
```diff
@@ -9,11 +9,20 @@ def __init__(self, _datetime):
         self._datetime = _datetime
         self._local_datetime = _datetime.astimezone(timezone('Europe/London'))
 
+    @property
+    def time_as_lang(self):
+        return {
+            '12:00AM': 'Midnight',
+            '12:00PM': 'Midday'
```
benthorner, line 16: \+ `3:30PM: 'Tea Time'`

tests/test_alert_date.py
```diff
@@ -19,7 +20,18 @@ def test_AlertDate_properties_work_with_bst():
...
     assert alerts_date.as_utc_datetime == sample_datetime
     assert alerts_date.as_local_datetime == datetime_in_bst
+
+
+@pytest.mark.parametrize('hour, minute, expected_lang', (
+    (0, 0, 'at midnight on Sunday 21 March 2021'),
+    (12, 0, 'at midday on Sunday 21 March 2021'),
+    (23, 59, 'at 11:59pm on Sunday 21 March 2021'),  # 12 hour clock
+))
+def test_AlertDate_at_midday_and_midnight(hour, minute, expected_lang):
```
benthorner, line 34: Minor / out-of-scope: we don't need the `AlertDate` prefix on these tests, since the whole file is now specific to the model.

idavidmcdonald (approved):
So generally looks good I think.

I'll approve it now for speed but I can't see any more uses of `relative-date` classes after this PR gets merged. Does this mean we can drop a bunch of related javascript too?

benthorner (approved):
Looks good 👍. Arguably these generic rules could be in the utils repo, but it looks like we're not applying them in any of our other apps, so they may as well go directly in here.

quis:
@idavidmcdonald `relative-date` is still used on the home page: https://github.com/alphagov/notifications-govuk-alerts/blob/main/templates/components/banner.html#L10

PR #89, opened 2021-05-25
Add expired date to past alerts

quis:
The content on past alert pages needs to make clear that the alert is not about a current emergency.

All we can say for certain is that the alert has stopped sending. The emergency itself may be ongoing.

We agreed to prioritise the date and time over the body of the alert, reversing the order of priority of information on current alerts.

---

image

---

Discussion: https://docs.google.com/document/d/1regPI_sAYQsHxUxgjQ3_nLiQ_E_qQkKiG4AIKATD0p8/edit#

### DIFF
--- a/src/alert.html
+++ b/src/alert.html
@@ -51,6 +51,11 @@
   </h1>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds alert-icon__container">
+      {% if alert_data.is_expired %}
+      <h2 class="govuk-heading-s">
+        Stopped sending <time datetime="{{ alert_data.expires_date.as_iso8601 }}">{{ alert_data.expires_date.as_lang }}</time>
+      </h2>
+      {% endif %}
       {{ alert_data.description | paragraphize }}
 
       <p class="govuk-body govuk-!-margin-bottom-0">


src/alert.html
```diff
@@ -51,6 +51,11 @@ <h1 class="govuk-heading-xl alerts-icon__container alerts-icon__container--48">
   </h1>
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds alert-icon__container">
+      {% if alert_data.is_expired %}
+      <h2 class="govuk-heading-s">
+        Ended <time datetime="{{ alert_data.expires_date.as_iso8601 }}">{{ alert_data.expires_date.as_lang }}</time>
```
benthorner, line 56: Is "ended" accurate here? For example, if the alert went:

```
Chemical leak in XYZ

Ended at ...
```

I'd read that as "the chemical leak has stopped", which may not be true. How about something more accurate, like "Broadcast ended at ...".

We use the term "broadcast" on the home page, so it seems reasonable to reuse it here.

quis: Yeah, I think you’re right. I’ve changed it to use the ‘Stopped sending’ version from Karl’s doc. This makes the language consistent because we say ‘Sent at’ beneath the alert text. I’ve amended the commit and added a screenshot to the description.

idavidmcdonald: To make it even more consistent, should we say 'Started sending' instead of 'Sent at' because it is continually sent at 60 second intervals? And someone might receive the alert 3 hours after it begins

benthorner: Thanks @quis. I think you missed an `at` in the code (comparing with the screenshot).

quis: The `at` is in the model:
https://github.com/alphagov/notifications-govuk-alerts/blob/eb35989cfc29971608304ba538bbf59f33797938/lib/alert_date.py#L25

benthorner: Ahh, I see, thanks for explaining. I think there's some merit in @idavidmcdonald's suggestion too, but this change is also good on its own.

benthorner (approved):

PR #91, opened 2021-05-25
More tests

idavidmcdonald:
https://www.pivotaltracker.com/story/show/178267667

Introduces some of the most basic tests to the jinja code for gov.uk/alerts

Review commit by commit

Note, this PR does not meet the aims of the pivotal ticket mentioned above. It instead lays the groundwork for it. The tests will need to come in a subsequent PR but I'd rather keep the PR small and iterate


benthorner (commented):
Great job comparing two approaches @idavidmcdonald 🏅. I've left a couple of comments. 

There's also a third approach you may want to consider: wrapping the majority of `build.py` in a function, which we can then call from the tests. Pros / cons / thoughts on this one:

- We can customise the source data (a file or a string) and output paths, so we can test what we want, but without the built pages overriding the real versions, which would be irritating for someone trying to look at them at the same time.
- Having the built pages also makes it easier to debug test failures, as the pages are actually visible. This is an extra advantage of the first approach as well, but we do manage without it in other apps.
- We would need to have some setup/teardown code to cleanup afterwards.
- We still get to test the majority of the code in `build.py`.
- Don't need to remember to run anything else beforehand.

Since this approach is kind of an extension of approach 1, we could also choose to pursue that for now, and then switch to this at a later date. On the other hand, it shouldn't be much effort to do this third approach now.

I think the second approach is good as well, for what we're trying to achieve with these tests. My order of preference of the approaches would be: 3, 1, 2. Do you want to get a third opinion?

tombye:
Just to say that my preference is for the third option @benthorner mentions, of wrapping most of `build.py` in a function so we can control the inputs, and so better assume the outputs.

tombye:
I think Ben talked about making the source data and output paths customisable as part of the design of the interface of the function we're exposing. This seems like something we still need to do, I'd assume would include most code below `if __name__ == '__main__':`

tests/src/__init__.py
```diff
@@ -0,0 +1,4 @@
+from build import setup_jinja_environment
+
+alerts = []
```
benthorner, line 3: Minor: rather than hard-code these here, we could maintain the convenience of a no-alerts environment by using a fixture instead.

```
# in tests/conftest.py
@pytest.fixture()
def env():
  return setup_jinja_environment([])
```

In individual tests, we can choose whether we want to add alerts.

```
def test_current_alerts_page(alert_dict, env):
  env.globals['alerts'] = [alert_dict]
  template = env.get_template("src/current-alerts.html")
  content = template.render()
  html = BeautifulSoup(content, 'html.parser')
  ...

def test_individual_alert_page(alert_dict, env):
  template = env.get_template("src/alert.html", {'alert_data': alert_dict})
  ...
```

build.py
```diff
@@ -10,27 +10,33 @@
...
 from lib.utils import DIST, REPO, ROOT, SRC, file_fingerprint, paragraphize
 
-jinja_loader = ChoiceLoader([
-    FileSystemLoader(str(REPO)),
-    PrefixLoader({
-        'govuk_frontend_jinja': PackageLoader('govuk_frontend_jinja')
-    })
-])
+
+def setup_jinja_environment(alerts):
```
benthorner, line 14: Minor: we could also encapsulate, and thus test, the code to decide the `target` for each kind of file. This isn't a big concern, though, as its covered by the smoke tests, with the exception of individual alert pages.

benthorner (approved):
Thanks for the chat earlier about this @idavidmcdonald and @tombye. I had another look at this and realised I misread it earlier: I think it's good to go.

I've left a couple of more minor comments, which @tombye may want to action in his follow-up work to make use of the foundation we've established here.

PR #92, opened 2021-05-26
Fix open graph image not showing due to relative path

idavidmcdonald:
At the moment if you see gov.uk/alerts shared on twitter we do not have
an open graph image for it.

Example: https://twitter.com/ollietheantijab/status/1397235543423979524

I believe the problem is that our open graph url is relative rather than
absolute.

https://stackoverflow.com/questions/9858577/open-graph-can-resolve-relative-url

By changing it to an absolute URL I think this will fix it and it will
start appearing.


benthorner (approved):
Nice catch 🎖️. 

This fix looks like it's worth a try: GOV.UK use an absolute URL as well.

PR #93, opened 2021-05-28
Add details of planned tests to the homepage and planned tests page

quis:
image
image

### DIFF
--- a/src/planned-tests.html
+++ b/src/planned-tests.html
@@ -40,16 +40,16 @@
       </h1>
 
       <h2 class="govuk-heading-m govuk-!-margin-top-6 govuk-!-margin-bottom-2">
-        Ongoing
+        Friday 18 June 2021
       </h2>
       <h3 class="govuk-body govuk-!-margin-bottom-6">
         Across the UK
       </h3>
       <p class="govuk-body">
-        Mobile phone networks in the UK are testing emergency alerts.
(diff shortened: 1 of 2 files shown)


src/planned-tests.html
```diff
@@ -67,14 +67,28 @@ <h2 class="govuk-heading-m govuk-!-margin-bottom-2">
...
       </h2>
       <h3 class="govuk-body govuk-!-margin-bottom-6">
-        Reading
+        Reading, Berkshire
       </h3>
       <p class="govuk-body">
         The UK government will send a test alert.
       </p>
       <p class="govuk-body">
         If you get an alert, your phone or tablet may make a loud siren-like sound. You do not need to take any action.
```
benthorner, line 79: I notice the alert above has an Android caveat. 

I'm guessing we expect all Apple devices to have updated by this time.

benthorner (approved):
📯

src/planned-tests.html
quis: Only Android devices can receive alerts sent on the test or operator channels.

Adoption of iOS 14.5 or above (i.e. iPhones that will receive alerts) is now at 76% (according to traffic to GOV.UK). I think this is high enough that we don’t need to caveat it.

benthorner: Ah, I didn't know that. Thanks for explaining.

PR #97, opened 2021-06-04
Reflect renamed class

quis:
I changed the name of this class in the CSS but forgot to update it in the HTML.


tombye (approved):
I also missed this when I reviewed those changes... 👍🏻

PR #98, opened 2021-06-04
Update opt out content for operator tests

karlchillmaid:
Small content update for the opt out page.

### DIFF
--- a/src/opt-out.html
+++ b/src/opt-out.html
@@ -64,7 +64,7 @@
         If you have an Android device, there’s a small chance you may get <a href="/alerts/planned-tests">test alerts</a> from your mobile network operator.
       </p>
       {% set testAlertButton %}
-        If you receive a test alert and want to opt out
+        If you want to opt out of mobile network operator alerts
       {% endset %}
       {% set testAlertAdvice %}
         <p class="govuk-body">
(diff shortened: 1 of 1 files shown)


src/opt-out.html
```diff
@@ -61,13 +61,10 @@ <h3 class="govuk-heading-s" id="mobile-network-tests">
...
       <p class="govuk-body">
-        Mobile network operators in the UK are <a href="/alerts/planned-tests">testing emergency alerts</a>.
-      </p>
-      <p class="govuk-body">
-        A small number of Android phones and tablets may receive these test alerts.
+        If you have an Android device, there’s a small chance you may get <a href="/alerts/planned-tests">test alerts</a> from your mobile network operator.
       </p>
       {% set testAlertButton %}
-        If you receive a mobile network operator test and want to opt out
+        If you receive a test alert and want to opt out
```
quis, line 67: I think it’s important to keep the word operator in here because the alert will be titled:
> ⚠️ Emergency Alert (Operator)

karlchillmaid, line 67:
```suggestion
        If you want to opt out of mobile network operator alerts
```

quis (approved):
Think this is a good improvement now 👍🏻

PR #99, opened 2021-06-04
Remove JS for govuk button

tombye:
While doing other work I noticed that [the JS for our buttons is wrong](https://github.com/alphagov/notifications-govuk-alerts/blob/main/src/assets/javascripts/govuk-frontend-details-and-button.mjs#L25). It searches for elements with a `govuk-button` class but then runs the JS for the details component.

The only reason this (my 😬) error hasn't caused any problems is that we don't have any buttons in our pages except the one for sharing and that isn't in the page when the GOVUK JS (linked to here) runs.

~So a useful outcome of this is that I don't think we need the JS at all. This removes it. More detailed reasons for why are in https://github.com/alphagov/notifications-govuk-alerts/pull/99/commits/30818930fe09b559b55ea0a476063118f35ea895.~

Edit: I don't think this is true any more, mainly based on https://github.com/alphagov/notifications-govuk-alerts/pull/99#issuecomment-855815846, so have made alternative changes in https://github.com/alphagov/notifications-govuk-alerts/pull/99/commits/69a5380784e23dd6612eced912af5945da0696a7 that, I hope, make more sense for this problem. This does mean the branch now has a silly name.


benthorner:
Nice cleanup 👍. I'm slightly concerned about this change, though: if we did have a link variant button in the future, are we confident we'd remember to add the JS back in for it?

On a more minor note, the `govuk-frontend-details.mjs` looks like it has a more general purpose than just initialising the details components, based on my reading of the comment at the top. How about we rename it to just `govuk-frontend.mjs`?

tombye:
> Nice cleanup 👍. I'm slightly concerned about this change, though: if we did have a link variant button in the future, are we confident we'd remember to add the JS back in for it?

I share this slight uneasiness and I kind of ignored it because of what seemed like the clarity of the argument for not including the JS for components not there. Coming back to that point, I still think it's sensible to not include it in the file for the details component because its name makes that clear what it contains (and doesn't). I'm less confident about leaving it out of the `sharing-button.mjs` code, because you need to know why it isn't there when we are adding a GOVUK button.

Because of that, and because I'd like to keep the name, I've changed the `sharing-button.js` code so it imports the code for the button component explicitly. I *think* this should let us keep the button code in the right file while also making it obvious that you need it if you're adding the HTML to your page.

> On a more minor note, the `govuk-frontend-details.mjs` looks like it has a more general purpose than just initialising the details components, based on my reading of the comment at the top. How about we rename it to just `govuk-frontend.mjs`?

It does look like it does more than just including the details component code, I think because its structure and comments are based on the file GOVUK Frontend uses to include all their components and match them to DOM nodes. I'd like to keep the `govuk-frontend-details.js` name to make it clear what it contains so I've rewritten its internals to only include code for the details component and to match the conventions use in other JS files.

In doing this work I also noticed the `/alerts/opt-out` page now uses a details component but doesn't include the JS so added it. See https://github.com/alphagov/notifications-govuk-alerts/pull/99/commits/60a1eb4a8c7fa19992a1ba55a6a086ed855e1bcd for more detail on why it was missed and what I suggest we do to prevent this happening in future.

src/assets/javascripts/govuk-frontend-details.mjs
```diff
@@ -1,30 +1,11 @@
-// JS Module used to combine all the JS modules used in the application into a single entry point,
-//
-// When processed by a bundler, this is turned into a Immediately Invoked Function Expression (IIFE)
-// The IIFE format allows it to run in browsers that don't support JS Modules.
-//
-// Exported items will be added to the window.GOVUK namespace.
-// For example, `export { initAll }` will assign `initAll` to `window.GOVUK.initAll`
-import { nodeListForEach } from 'govuk-frontend/govuk/common'
```
benthorner: Minor: how come we're not using `nodeListForEach` anymore? The new code is longer, so I'm not sure what we gain from the change.

benthorner (approved):
Thanks for the thorough explanations @tombye 💯. I can see what you mean about the `-init.mjs -> .js` approach now being consistent for the button and the details components.

>  I'd like to keep the govuk-frontend-details.js name

It took me a while to understand a subtlety here: we want to have our JS as fine-grained as possible, so we only run what we need on each page, to improve load times. Knowing that, the changes make even more sense 👍.

src/assets/javascripts/govuk-frontend-details.mjs
tombye: Looking at [the nodeListForEach code](https://github.com/alphagov/govuk-frontend/blob/4b5ed0ada3b64722ae2074dfa0db5d4d3a45cb79/src/govuk/common.js#L6), it adds the following to the native looping I use:
1. use of `nodeList.forEach` if it exists
2. uses `.call` to bind the `this` in your callback function to the current node in the list

`nodeList.forEach` is probably a bit faster but we don't use `this` so the benefits seemed a bit minor for the extra code (`nodeListForEach` is actually bit longer than my loop) and addition of a dependency. This is total opinion though and interested to know if you think my version is the right option here.

tombye: With some hindsight, I think this change adds noise to the history and adds only marginal gains so I've reverted back to `nodeListForEach` in https://github.com/alphagov/notifications-govuk-alerts/pull/99/commits/6fe6012517ce915a4e714571acf7dc86aa103ccc which, I think resolves this (do say if you don't agree).

benthorner: That's OK. I don't mind much either way, and was just surprised / interested by the difference. Since the file is doing stuff with govuk-frontend, I lean towards using their tools, so we're more consistent.

Thanks for changing it back - I was just slow to respond, so feel free to merge with/out that commit.

PR #104, opened 2021-06-08
Add content about current and past alerts

karlchillmaid:
The PR adds content about (and links to) current and past alerts on GOV.UK.

We removed this line when we launched the MVP site, but now the current and past alerts pages are live we can reinstate it.

### DIFF
--- a/src/when-you-get-an-alert.html
+++ b/src/when-you-get-an-alert.html
@@ -72,7 +72,7 @@
         If you want to see an alert again
       </h2>
       <p class="govuk-body">
-        Visit gov.uk/alerts to see <a href="/alerts/current-alerts">current alerts</a> and <a href="/alerts/current-alerts">past alerts</a>.
+        You can find <a href="/alerts/current-alerts">current alerts</a> and <a href="/alerts/current-alerts">past alerts</a> at gov.uk/alerts.
       </p>
       <p class="govuk-body">
         You can also find them on your phone or tablet. 
--- a/src/when-you-get-an-alert.html
+++ b/src/when-you-get-an-alert.html
@@ -75,7 +75,7 @@
         Visit gov.uk/alerts to see <a href="/alerts/current-alerts">current alerts</a> and <a href="/alerts/current-alerts">past alerts</a>.
       </p>
       <p class="govuk-body">
-        You can also find current and past alerts on your phone or tablet. 
+        You can also find them on your phone or tablet. 
       </p>
       {% set androidButton %}
         <span class="govuk-visually-hidden">Find alerts on </span>Android phones and tablets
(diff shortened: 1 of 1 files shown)


src/when-you-get-an-alert.html
```diff
@@ -72,7 +72,10 @@ <h2 class="govuk-heading-m">
         If you want to see an alert again
       </h2>
       <p class="govuk-body">
-        You can find current and past alerts on your phone or tablet.
+        Visit gov.uk/alerts to see a list of <a href="/alerts/current-alerts">current alerts</a> and <a href="/alerts/current-alerts">past alerts</a>.
```
karlchillmaid, line 75:
```suggestion
        Visit gov.uk/alerts to see <a href="/alerts/current-alerts">current alerts</a> and <a href="/alerts/current-alerts">past alerts</a>.
```

src/when-you-get-an-alert.html
```diff
@@ -72,7 +72,10 @@ <h2 class="govuk-heading-m">
...
+      </p>
+      <p class="govuk-body">
+        You can also find current and past alerts on your phone or tablet.
```
karlchillmaid, line 78:
```suggestion
        You can also find them on your phone or tablet. 
```

sakisv (approved):

src/when-you-get-an-alert.html
```diff
@@ -72,7 +72,10 @@ <h2 class="govuk-heading-m">
         If you want to see an alert again
       </h2>
       <p class="govuk-body">
-        You can find current and past alerts on your phone or tablet.
+        Visit gov.uk/alerts to see <a href="/alerts/current-alerts">current alerts</a> and <a href="/alerts/current-alerts">past alerts</a>.
```
karlchillmaid, line 75: Does it sound weird to tell users to visit a site they’re already on?

It’s less active, but might be better to say:

```suggestion
        You can find <a href="/alerts/current-alerts">current alerts</a> and <a href="/alerts/current-alerts">past alerts</a> at gov.uk/alerts.
```

src/when-you-get-an-alert.html
```diff
@@ -72,7 +72,10 @@ <h2 class="govuk-heading-m">
...
+      </p>
+      <p class="govuk-body">
+        You can also find them on your phone or tablet.
```
karlchillmaid, line 78: If we change line 75, we should update this to:

```suggestion
        You can also search for them on your phone or tablet. 
```

src/when-you-get-an-alert.html
idavidmcdonald: When I read this I did find it a bit weird. I prefer your suggestion

karlchillmaid: Thanks @idavidmcdonald.

Normally we wouldn’t mention the URL at all – but this content is trying to do 2 things:

* signpost the current and past alerts pages
* teach users what to do in the future

idavidmcdonald (approved):

PR #105, opened 2021-06-09
Add JS tests for sharing button

tombye:
Adds tests for the sharing button JS.

Part of https://www.pivotaltracker.com/story/show/177597156. This should complete the 'test the JavaScript works' task because we are not testing GOVUK Frontend code.

To run these tests only, use this command:

`npx jest --config tests/javascripts/jest.config.js tests/javascripts/sharing-button.test.js`

This pull request also includes:
- adding checks for copy-to-clipboard support instead of assuming it exists if the Web Share API doesn't
- updates the RegExp used to trim the URL string to use the `String.prototype.trim` polyfill from MDN

### DIFF
--- a/src/assets/javascripts/sharing-button.mjs
+++ b/src/assets/javascripts/sharing-button.mjs
@@ -7,11 +7,14 @@ export default function sharingButton () {
   var urlString
   var onClick
 
+  // return early if no of the required APIs are available
+  if (!navigator.share && !document.queryCommandSupported('copy')) return
+
   btn.className = 'govuk-button govuk-button--secondary'
   btn.setAttribute('data-module', 'govuk-button')
   if (navigator.share) {
     btn.innerHTML = 'Share link'
-  } else {
+  } else if (document.queryCommandSupported('copy')) {
     btn.innerHTML = 'Copy link'
   }
 
--- a/src/assets/javascripts/sharing-button.mjs
+++ b/src/assets/javascripts/sharing-button.mjs
@@ -19,7 +19,8 @@ export default function sharingButton () {
   }
 
   if (navigator.share) {
-    urlString = url.innerHTML.replace(/^[\s]+$/, '')
+    // Regexp taken from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim#polyfill
+    urlString = url.innerHTML.replace(/^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, '')
 
     onClick = function (evt) {
       try {
--- a/tests/javascripts/sharing-button.test.js
+++ b/tests/javascripts/sharing-button.test.js
@@ -1,0 +1,195 @@
+import helpers from './support/helpers.js'
+import sharingButton from '../../src/assets/javascripts/sharing-button.mjs'
+
+const URL = 'https://www.gov.uk/alerts/3-may-2021'
+const htmlFixture = `
+      <div class="share-url">
+        <p class="govuk-body">
+          ${URL}
+        </p>
+      </div>`
+
+afterAll(() => {
+  // clear up methods in the global space
+  document.queryCommandSupported = undefined
+})
+
+afterEach(() => {
+  document.body.innerHTML = ''
+
+  jest.restoreAllMocks()
+})
+
+describe('If the Web Share API is supported', () => {
+
+  beforeEach(() => {
+
+    navigator.share = jest.fn(() => {})
+
+    // copy command would be supported if navigator.share is
+    document.queryCommandSupported = jest.fn(command => command === 'copy')
+
+    document.body.innerHTML = htmlFixture
+
+    sharingButton()
+
+  })
+
+  afterEach(() => {
+
+    delete navigator.share
+
+  })
+
+  describe('On page load', () => {
+
+    test('A button should be added below the URL to be shared', () => {
+
+      const URLParagraph = document.querySelector('.share-url > .govuk-body')
+
+      expect(URLParagraph.nextElementSibling).not.toBe(null)
+      expect(URLParagraph.nextElementSibling.classList.contains('govuk-button')).toBe(true)
+
+    })
+
+    test('The button should be for sharing the link', () => {
+
+      const button = document.querySelector('.share-url > .govuk-button')
+
+      expect(button.textContent.trim()).toEqual('Share link')
+
+    })
+
+  })
+
+  describe('If the button is clicked', () => {
+
+    test('The correct information should be sent to the Web Share API', () => {
+
+      const button = document.querySelector('.share-url > .govuk-button')
+      const clickEvent = new window.MouseEvent(
+        'click',
+        {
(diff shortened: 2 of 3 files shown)


tests/javascripts/sharing-button.test.js
```diff
@@ -0,0 +1,195 @@
...
+        </p>
+      </div>`
+
+afterAll(() => {
+  // clear up methods in the global space
+  document.queryCommandSupported = undefined
+})
+
+afterEach(() => {
+  document.body.innerHTML = ''
```
benthorner, line 18: Minor: we don't do this in the Admin app, do we need it in these tests?

tests/javascripts/sharing-button.test.js
```diff
@@ -0,0 +1,195 @@
...
+const htmlFixture = `
+      <div class="share-url">
+        <p class="govuk-body">
+          ${URL}
+        </p>
+      </div>`
+
+afterAll(() => {
+  // clear up methods in the global space
+  document.queryCommandSupported = undefined
```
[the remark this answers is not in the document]
benthorner: Also, could we do `delete` instead of `= undefined` (similarly for other assignments below)? My assumption is that `delete` is more realistic, and in any case it's helpful if we can be consistent.

tests/javascripts/support/helpers.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+}
+
+// Very basic class for stubbing the Range interface
+// Only contains methods required for current tests
+class RangeMock extends DOMInterfaceMock {
+
+  constructor (jest) {
+    super(jest, { props: [], methods: ['selectNodeContents', 'setStart'] });
```
benthorner, line 33: `setStart` isn't used anywhere, do we need it?

src/assets/javascripts/sharing-button.mjs
```diff
@@ -7,16 +7,20 @@ export default function sharingButton () {
   var urlString
   var onClick
 
+  // return early if no of the required APIs are available
+  if (!navigator.share && !document.queryCommandSupported('copy')) return
```
benthorner, line 11: Minor: these looks like they belong in the previous commit.

tests/javascripts/support/helpers.js
```diff
@@ -0,0 +1,49 @@
+// helpers for mocking DOM interfaces
+// see https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model#DOM_interfaces
+
+// Base class for mocking an DOM API interfaces not in JSDOM
+class DOMInterfaceMock {
```
benthorner, line 5: Nice abstraction!

benthorner (approved):
Looks good to me 👍. 

I've left some more minor comments you could address. In particular, I think it would help future reviews if we can structure `beforeEach`s like we do in the Admin app, so there's a bit less to think about during the review.

src/assets/javascripts/sharing-button.mjs
tombye: Agreed and addressed in https://github.com/alphagov/notifications-govuk-alerts/pull/105/commits/f1f3493e607847e8d7b738af605c811eacbe1415

tests/javascripts/support/helpers.js
tombye: No, it's used in version of this mock used in the admin app, which also has to support tests that alter the position of the range. Removed in https://github.com/alphagov/notifications-govuk-alerts/pull/105/commits/f1f3493e607847e8d7b738af605c811eacbe1415.

tests/javascripts/sharing-button.test.js
tombye: I think we do it in quite a few places in the admin app (see https://github.com/alphagov/notifications-admin/search?q=document.body.innerHTML) but I might be missing your point here about its value in these tests maybe. What do you think?

tests/javascripts/sharing-button.test.js
```diff
@@ -0,0 +1,195 @@
...
+
+    delete navigator.share
+
+  })
+
+  describe('On page load', () => {
+
+    test('A button should be added below the URL to be shared', () => {
+
+      const URLParagraph = document.querySelector('.share-url > .govuk-body')
```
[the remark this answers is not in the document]
tombye: Agreed. I often struggle with naming camelCase variables based on acronyms tbh so happy to go with your suggestion (in https://github.com/alphagov/notifications-admin/search?q=document.body.innerHTML).

tests/javascripts/sharing-button.test.js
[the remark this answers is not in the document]
tombye: Sure, I borrowed this line from [a similar test in the admin app](https://github.com/alphagov/notifications-admin/blob/2871d4cb0433964b2e96f07644aee314b6d576c3/tests/javascripts/apiKey.test.js#L8) where, I think, `delete` isn't used elsewhere so there isn't the same requirement for consistency. The form including an assignment is something I picked up a while ago, which I think is intended to be less ambiguous than `delete` (because, in JavaScript you can often be wrong about what an operation will perform :-S). Happy to convert to use `delete` here for consistency (done in https://github.com/alphagov/notifications-govuk-alerts/pull/105/commits/f1f3493e607847e8d7b738af605c811eacbe1415).

tests/javascripts/sharing-button.test.js
```diff
@@ -0,0 +1,195 @@
+import helpers from './support/helpers.js'
+import sharingButton from '../../src/assets/javascripts/sharing-button.mjs'
+
+const URL = 'https://www.gov.uk/alerts/3-may-2021'
+const htmlFixture = `
```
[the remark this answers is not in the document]
tombye: Agreed, and changed in https://github.com/alphagov/notifications-govuk-alerts/pull/105/commits/f1f3493e607847e8d7b738af605c811eacbe1415.

tests/javascripts/sharing-button.test.js
benthorner: Ah, fair enough - we're looking in different places for our precedents. More objectively, I'd assumed that, like the module state, JSDOM was being reset for each test module. However, [it appears I'm mistaken](https://stackoverflow.com/questions/42805128/does-jest-reset-the-jsdom-document-after-every-suite-or-test).

In practice, we're likely to do an assignment to `innerHTML` if we're testing something else that makes use of it. On the other hand, it's good practice to clean up. I don't mind if we keep it, as it's sort of optional.

tests/javascripts/sharing-button.test.js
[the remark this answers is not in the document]
benthorner: Oh, interesting: I would say `delete` is less ambiguous than `undefined`. Aside from consistency, using `delete` means the object is restored back to what it was, whereas `= undefined` means it still has the property declared on it. I doubt it matters much either way, so just being consistent in the same file is fine with me.

benthorner:
Thanks for looking at my comments @tombye! 

I think this is good to go, if you'd like to merge it now.

PR #106, opened 2021-06-09
Fix fallback image for alert icon

tombye:
I noticed that the alert icon image, used as a fallback for the SVG version, is broken in IE8 (which uses it due to lack of support for SVG).

### Current

past_alerts_current_ie8

### Fixed

past_alerts_new_ie8

### Current

alert_current_ie8

### Fixed

alert_new_ie8

Looks like the fallback image has classes intended for the example image on the home page (of the phone). This fixes it to instead have the same classes as the SVG it replaces.

### DIFF
--- a/templates/components/alerts_icon.html
+++ b/templates/components/alerts_icon.html
@@ -16,7 +16,7 @@
   {% if degrade %}
     <!--<![endif]-->
     <!--[if IE 8]>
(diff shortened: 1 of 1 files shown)


benthorner (approved):
Nice catch - I can see these are the same as the non-IE8 classes 👍.

PR #107, opened 2021-06-09
Update instructions for drivers

karlchillmaid:


src/when-you-get-an-alert.html
```diff
@@ -63,10 +66,16 @@ <h1 class="govuk-heading-xl">
...
-        If you’re driving when you get an alert
+        If you’re driving or riding when you get an alert
       </h2>
       <p class="govuk-body">
-        Find somewhere safe to stop before using your phone or tablet. It is illegal to use a hand-held device while driving.
+        Find somewhere safe to stop before using your phone or tablet. 
+      </p>
+      <!--<p class="govuk-body">
+        If you cannot find a place to stop, turn on your radio and wait for the news.
+      </p>-->
```
karlchillmaid, line 76: Spoke to DfT and we agreed not to include this (for now).

karlchillmaid, line 76:
```suggestion
```

src/when-you-get-an-alert.html
```diff
@@ -63,10 +66,13 @@ <h1 class="govuk-heading-xl">
         <p>The government and mobile phone networks are <a href="/alerts/planned-tests">testing emergency alerts</a>. You may get an alert if you live in, or travel through, a test area.</p>
       </div>
       <h2 class="govuk-heading-m">
-        If you’re driving when you get an alert
+        If you’re driving or riding when you get an alert
```
benthorner, line 69:
```suggestion
        If you’re travelling when you get an alert
```

We've set "travel" above, so can we say it here?

src/when-you-get-an-alert.html
```diff
@@ -53,6 +53,9 @@ <h1 class="govuk-heading-xl">
           read out the alert
         </li>
       </ul>
+      <p class="govuk-body">
+        The sound and vibration will last for 10 seconds.
```
benthorner, line 57: How do we know this?

src/when-you-get-an-alert.html
karlchillmaid: Travelling is not the same as driving or riding.

If you’re travelling – on a train or as a passenger in a motor vehicle – there’s nothing to stop you checking your device.

The same is not true for anyone responsible for driving/riding a vehicle.

src/when-you-get-an-alert.html
karlchillmaid: I checked this with Pete D at Fujitsu.

After reviewing the video from the Sizewell test, he confirmed that all phones stopped making the sound after approx 10 seconds.

I’ll add ‘about’ to ~remove~ _include_ a little ambiguity. ;-)

karlchillmaid:
```suggestion
        The sound and vibration will last for about 10 seconds.
```

PR #110, opened 2021-06-15
Test local links in pages

tombye:
Test that all links to `/alerts` pages are correct. Part of this story:

https://www.pivotaltracker.com/story/show/177597156

### DIFF
--- a/tests/test_local_links.py
+++ b/tests/test_local_links.py
@@ -1,0 +1,87 @@
+import re
+from datetime import datetime
+from glob import glob
+from pathlib import Path
+from urllib.parse import urlparse
+
+import pytest
+import pytz
+from bs4 import BeautifulSoup
+from freezegun import freeze_time
+
+from lib.alerts import Alerts
+
+
+def get_page_html(env, template_path, alert_data=None):
+    template = env.get_template(template_path)
+    if alert_data is not None:
+        content = template.render({'alert_data': alert_data})
+    else:
+        content = template.render()
+    return BeautifulSoup(content, 'html.parser')
+
+
+def get_local_route_from_template_path(template_path):
+    local_route = '/alerts/' + Path(template_path).stem
+    if local_route != '/alerts/index':
+        return local_route
+    else:  # index page is mapped to root of /alerts directory
+        return '/alerts'
+
+
+all_templates = glob('./src/*.html')
+all_templates_except_alerts = filter(lambda template: template != './src/alert.html', all_templates)
+local_routes_except_alerts = [
+    get_local_route_from_template_path(template_path)
+    for template_path in all_templates]
+
+
+def test_local_links_in_pages_with_no_alerts(env):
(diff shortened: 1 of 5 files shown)


tests/test_local_links.py
```diff
@@ -0,0 +1,87 @@
...
+    template = env.get_template(template_path)
+    if alert_data is not None:
+        content = template.render({'alert_data': alert_data})
+    else:
+        content = template.render()
+    return BeautifulSoup(content, 'html.parser')
+
+
+def get_local_route_from_template_path(template_path):
+    local_route = '/alerts/' + Path(template_path).stem
```
CrystalPea, line 25: what does `.stem` do? :3

tests/test_local_links.py
```diff
@@ -0,0 +1,87 @@
...
+        html = get_page_html(env, re.sub(r'^\./{1}', '', template_path))
+        local_links = html.select("a[href^='/alerts']")
+
+        if len(local_links) == 0:
+            assert True
+            return
+
+        for link in local_links:
+            path = urlparse(link['href']).path
+            assert path in local_routes_except_alerts
```
CrystalPea, line 50: I feel this (and next) test doesn't fully explain what it tries to do. The test name could be more descriptive. Maybe something like: `test_local_links_in_pages_with_no_alerts_lead_to_expected_routes` or something? What do you think?

We could also add a comment above `if len(local_links) == 0:` explaining that some pages won't have links (what pages are those?). What do you think?

CrystalPea (changes requested):
Hey, well done on adding these tests, the setup looks a bit gnarly, well done for figuring it out 👏🏼 . The names could be more descriptive, otherwise this is good to go ☺️

tests/test_local_links.py
tombye: I see what you mean. I tried a few different versions for the test names and settled on the changes in https://github.com/alphagov/notifications-govuk-alerts/pull/110/commits/63c0bfaa40b9ef9f8a83799f357cb511cf5bc953. Do you think that reads any better?

tests/test_local_links.py
tombye: It give you the filename with its suffix (see https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem). I was using a RegExp to do the same but see that we use `Path` in other places so thought this would be easier.

CrystalPea: You mean without the suffix, right? At least that's what the docs say. Thanks for the link, I now get it ☺️

tests/test_local_links.py
CrystalPea: Much better 👏🏼

CrystalPea (approved):

tests/test_local_links.py
tombye: I do! 😬

PR #111, opened 2021-06-15
Add test for inline script

tombye:
We use the GOVUK Frontend library template.html
file as the base layout of our pages. It includes
an inline script which our Content Security Policy
allows by referencing a sha256 of its contents.

This adds a test to check the script's contents
match the sha256. If this test fails, the sha256
has changed and so the sha256 in our Content
Security Policy will need to be updated to the new
value.

Part of this story:

https://www.pivotaltracker.com/story/show/177597156

The Content Security Policy is added here:

https://github.com/alphagov/notifications-broadcasts-infra/blob/main/terraform/modules/govuk-alerts-website/files/main.py

It is tested here:

https://github.com/alphagov/notifications-broadcasts-infra/blob/main/terraform/modules/govuk-alerts-website/files/test_inject_headers.py

Both contain references to the sha256.

### DIFF
--- a/tests/test_csp_sha.py
+++ b/tests/test_csp_sha.py
@@ -1,0 +1,27 @@
+import base64
+import hashlib
+
+from bs4 import BeautifulSoup
+
+from tests.src import env
+
+
+def test_csp_sha():
+    script_sha256_base64_encoded = b"+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU="
+    env.filters['file_fingerprint'] = lambda path: path  # stub out filters
+
+    template = env.get_template('src/index.html')
+    content = template.render()
+    html = BeautifulSoup(content, 'html.parser')
+    inline_script = html.select_one('body > script')
+
+    assert inline_script is not None
+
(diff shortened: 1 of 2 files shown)


tests/test_csp_sha.py
```diff
@@ -0,0 +1,27 @@
+import base64
+import hashlib
+
+from bs4 import BeautifulSoup
+
+from tests.src import env
+
+
+def test_csp_sha():
```
CrystalPea, line 9: we could name it `test_content_security_policy_sha`, both in filename and test name, this way it will be more readable and provide additional context for a person that reads it for the first time, or first time in a long while ☺️

CrystalPea (changes requested):
Looks good, and it fails if I change the SHA, which is good. I would suggest changing the test file and test name though, to reduce acronym burden.

tests/test_csp_sha.py
tombye: Good point 💯. I've made those changes in https://github.com/alphagov/notifications-govuk-alerts/pull/111/commits/971ff6044dd0e8a86e90bf5aa4cbe2a27817051f.

PR #112, opened 2021-06-17
Fix date of planned test

quis:
It’s actually today

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -44,7 +44,7 @@
   {{ banner(alerts.current | length, alerts.last_updated_date) }}
   {% else %}
     <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
-    {{ planned_tests_banner(2, "Thursday 17 June and Tuesday 29 June") }}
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -44,7 +44,7 @@ <h1 class="govuk-heading-xl">
   {{ banner(alerts.current | length, alerts.last_updated_date) }}
   {% else %}
     <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
-    {{ planned_tests_banner(2, "Friday 18 June and Tuesday 29 June") }}
+    {{ planned_tests_banner(2, "Thursday 17 June and Tuesday 29 June") }}
```
benthorner, line 47: Should we say Thursday **and** Friday?

quis: If there’s more than two I think it’s less cluttered to give a range. This also means we need to add Friday’s test to the planned tests page. I’ve done that in 554ba4b

benthorner (approved):

PR #118, opened 2021-06-18
Put hashes in asset filenames, not querystrings

quis:
Using querystring hashes is a DoS vulnerability because it means that a request with a new querystring will miss the edge and pass through to the origin.


rjbaker (approved):
Looks good. I updated the file_fingerprint test to match the change.

gulpfile.js
```diff
@@ -134,6 +144,10 @@ const javascripts = {
   }
 };
 
+const hashJavascriptFilenames = () => src(paths.dist + 'javascripts/*.js')
```
benthorner: Minor: this is inconsistent with the style we have for `scss` below, which has a `compile` task. Could we have made it a `hashFilenames` key in the existing `javascripts` const, rather than adding a new one?

gulpfile.js
```diff
@@ -14,6 +14,8 @@ plugins.sass = require('gulp-sass');
 plugins.gulpStylelint = require('gulp-stylelint');
 plugins.gulpif = require('gulp-if');
 plugins.postcss = require('gulp-postcss');
+plugins.hash = require('gulp-hash-filename');
```
benthorner, line 17: This is subtle: we're now relying on the hash generated by this library being the same as the one we're generating in Python. While I can [see that it is](https://github.com/intervalia/gulp-hash-filename/blob/master/lib/generateHash.js), it's worth at least a code comment to explain the two need to be kept in sync.

gulpfile.js
```diff
@@ -144,10 +146,26 @@ const javascripts = {
...
 
-const hashJavascriptFilenames = () => src(paths.dist + 'javascripts/*.js')
+const hashJavascriptFilenames = () => src([
+  paths.dist + 'javascripts/*.js',
+  paths.dist + 'javascripts/*.js.map'
+])
   .pipe(plugins.hash())
   .pipe(dest(paths.dist + 'javascripts/'));
 
+const fixSourcemapFilenames = () => src(paths.dist + 'javascripts/*.js-*.map')
```
benthorner, line 156: Rather than use a plugin and have to hack around it, could we just write own our JS instead? That should be no longer than the length of this.

benthorner (commented):
I appreciate this has gone in already, but I've got some concerns about the approach we've taken here, which I'm finding quite subtle and pretty hard to follow. Was this an urgent change, or could it have waited a day or two? I appreciate this isn't my field of expertise, so I'd be interested to hear @tombye's thoughts on the approach as well.

quis:
This is 90% working now. Just need to figure out why `filenameFormat` is being ignored in this one place:
https://github.com/alphagov/notifications-govuk-alerts/pull/118/files#diff-25789e3ba4c2adf4a68996260eb693a441b4a834c38b76167a120f0b51b969f7R52

templates/_layout.html
```diff
@@ -31,7 +31,7 @@
 
   {# For older browsers to allow them to recognise HTML5 elements such as `<header>` #}
   <!--[if lt IE 9]>
-    <script src="/alerts/assets/javascripts/vendor/html5shiv/html5shiv.min.js"></script>
+    <script src="/alerts/assets/javascripts/vendor/html5shiv/html5shiv.min-init.js"></script>
```
quis, line 34:
```suggestion
    <script src="/alerts/assets/javascripts/vendor/html5shiv/html5shiv.min.js"></script>
```

lib/utils.py
```diff
@@ -20,4 +21,6 @@ def paragraphize(value, classes="govuk-body-l govuk-!-margin-bottom-4"):
 
 def file_fingerprint(path, root=DIST):
     contents = open(str(root) + path, 'rb').read()
-    return path + '?' + hashlib.md5(contents).hexdigest()
+    hash = hashlib.sha256(contents).hexdigest()[:8]
```
quis, line 24:
```suggestion
    # Must be the same algorithm and format used by the tasks in gulpfile.js
    hash = hashlib.sha256(contents).hexdigest()[:8]
```

benthorner:
Just in case it helps: would it be easier to get the Python code to do the renaming? It already knows where the compiled assets are, and how to fingerprint them, so it's not a big step to rename them as well.

rjbaker:
@benthorner Totally forgot that Ollie had a go at doing it in Python: https://github.com/OllieJC/notifications-govuk-alerts/commit/6126d5e246055bc04ad2b700e683eb0093c758d7

tombye:
The only problem with doing it all in python is that the JavaScript build generates a sourcemap to give the browser a view of the uncompressed code. This is linked to the compressed code by a comment at it's end, containing the name of the sourcemap. In other words, if we rename the compressed file, we need to rename this too. If rollup (the library that builds the JS) is allowed to insert the hash itself, it will update this reference too.

All that being the case, we've been looking at how we can keep the hash generation in the frontend build but make it work with the python. Caveat: we could just do a search-and-replace on the sourcemap comment but messing with the contents of files doesn't seem a great idea if we can avoid it.

benthorner:
Thanks for your thoughts @tombye. 

I've wondered about source maps in the past: why do we have them? If they're just a nice-to-have, then we could just get rid of them instead of spending more effort to make them work.

tombye:
> I've wondered about source maps in the past: why do we have them? If they're just a nice-to-have, then we could just get rid of them instead of spending more effort to make them work.

It can be very hard to debug JS in the browser without them as we obfuscate our code as well as minify it. They aren't necessarily needed on production but I can't seem much value in doing extra work to remove them when:
- they aren't loaded until devtools is open
- we might need to debug JS on production at some point

tombye:
Me and @quis have been working on some changes which move this to the approach I mentioned in https://github.com/alphagov/notifications-govuk-alerts/pull/118#issuecomment-872044985.

This new approach:
- leaves the process of adding SHAs to the static asset filenames to the Gulp tasks
- changes the file_fingerprint filter from finding the SHA for filenames from their contents to finding the SHA'ed filename through a regex, built from the original filename

That means as long as assets have 8 character SHAs in their name, we trust Gulp to generate the right SHA for their contents.

This also includes:
- some tidying up of asset URLs across the pages
- passing all assets we can through `file_fingerprint`
- extra logic to ignore a small set of assets we can't add SHAs to because we can't change their URLs

benthorner:
Thanks for the update @tombye, that does sound like a simpler approach: all the knowledge about the hashing is in one place, and the Python code just needs to care about knowing what the file names are (a simpler goal).

I noticed there are quite a few "WIP" commits. Can we squash those at all? That might make the PR easier to review.

tombye:
> I noticed there are quite a few "WIP" commits. Can we squash those at all? That might make the PR easier to review.

I went through those commits, squashed them into one and added a summary of the combined changes. Hopefully this should help.

gulpfile.js
```diff
@@ -69,7 +91,8 @@ const javascripts = {
       ]
     }).then(bundle => {
       return bundle.write({
-        file: paths.dist + 'javascripts/govuk-frontend-details.js',
+        dir: paths.dist + 'javascripts/',
+        entryFileNames: '[name]-[hash].js',
```
benthorner, line 95: It took me a long time to find the code that generates `[hash]`, to be sure it will always be 8 chars long. Would have been helpful to have a reference in the commit message that introduces this: https://github.com/rollup/rollup/search?q=digest

gulpfile.js
```diff
@@ -98,7 +123,8 @@ const javascripts = {
       ]
     }).then(bundle => {
       return bundle.write({
-        file: paths.dist + 'javascripts/sharing-button.js',
+        dir: paths.dist + 'javascripts/',
+        entryFileNames: '[name]-[hash].js',
```
benthorner, line 127: Minor: could we use a loop to generate this hash for each of the 3 input files? At the moment we have a lot of duplication (each one entry is the same).

Something like this:

```
const javascripts = Object.fromEntries(
  ['sharing-button-init.mjs', ...].map((name) => {
    [name, () => {
      rollup.rollup(...
```

benthorner (approved):
Thanks for trying @tombye ⭐. 

I ended up looking at the overall diff: from a reviewers perspective, the commits are hard to follow because they tell a meandering story of dead ends and small fixes (stuff I don't need to know). Fortunately the overall diff wasn't too hard to look at in one go. Definitely something I'll be pushing for in future PRs I review, though. As a quick suggestion, I reckon you could squash all but the last two commits into one that explains the approach.

Anyway, nice one for pursuing this and finding a single approach for it, especially with the weird rollup `[name]-[hash]` bit 💯. I've only got a few more minor comments, so I'll approve this now.

gulpfile.js
quis: I’ve done something similar to your suggestion in aadec78

benthorner: Nice 👍

PR #127, opened 2021-06-25
Add survey link to the ‘Reasons you might get an alert’ page

karlchillmaid:
We’ve been asked to make the link to the survey more prominent for the Reading test.

The ‘Reasons you might get an alert’ page seems like the most appropriate place to do this.


quis (approved):
image

PR #136, opened 2021-07-02
Test details component has JS

tombye:
Because the `<details>` tag is supported natively in all modern browsers it is hard to know when adding the GOVUK Frontend details component if you've forgetten to add the JS for it.

As well as polyfilling `<details>` in older browsers, the JS also adds the correct aria roles, both of which are important but hard to spot when testing the visual design in Chrome/Firefox.

This adds a check for this scenario, to ensure all the code needed is present.

Part of https://www.pivotaltracker.com/story/show/177597156


CrystalPea (commented):
Hey, would it be useful to add to the commit message what those details are that we are testing, why we need them, and what happens if we don't have this accompanying JS script?

I'm asking because I have no idea and would find it helpful in the future if looking at the commits as documentation, and maybe other people would find it helpful, too?

tombye:
> Hey, would it be useful to add to the commit message what those details are that we are testing, why we need them, and what happens if we don't have this accompanying JS script?

I've rewritten the last commit to include this information. It should help explain it better (and provide a place for that information to live) but do say if it doesn't make sense or I've missed anything.

CrystalPea (approved):
Thanks, that additional context in the 2nd commit message really helped!

PR #140, opened 2021-07-13
Add 14 July operator test to site

quis:
Should be deployed end of the day on 13 July.

***

image

***

image


benthorner (approved):
I wonder when we'll feel the need to automate this...

PR #142, opened 2021-07-15
Bump GOVUK Frontend Jinja

tombye:
We bumped [GOVUK Frontend](https://github.com/alphagov/govuk-frontend) (the library that provides the CSS, JavaScript and images) in https://github.com/alphagov/notifications-govuk-alerts/pull/113 so I thought we should also bump [GOVUK Frontend Jinja](https://github.com/LandRegistry/govuk-frontend-jinja/) (the library that provides the Jinja templates and components to render HTML).

They've both moved a minor version on since then so this bumps both libraries to the latest (v3.13.0).

I figured that, while I was doing this work I should also see what else needs updating (via `npm outdated`). Based on that, I bumped all the other NPM modules we use to their latest versions. I also made changes to our code needed by these new versions.

Run `make bootstrap` or `npm ci` to bring down all the updates.

### GOVUK Frontend Jinja

[GOVUK Frontend Jinja](https://github.com/LandRegistry/govuk-frontend-jinja/) has its own version numbers but each version is based on a version of [GOVUK Frontend](https://github.com/alphagov/govuk-frontend). So this bumps [GOVUK Frontend Jinja](https://github.com/LandRegistry/govuk-frontend-jinja/) to v1.4.0 which supports [GOVUK Frontend](https://github.com/alphagov/govuk-frontend) v3.13.0. All this is listed in [their CHANGELOG](https://github.com/LandRegistry/govuk-frontend-jinja/blob/main/CHANGELOG.md).

### Local changes

[Jest](https://jestjs.io/) now requires you to specify the environment our JavaScript tests run in to be [JSDOM ](https://github.com/jsdom/jsdom)(a simulation of the Document Object Model) so this changes our Jest config to specify this.

[Gulp-sass](https://github.com/dlmanning/gulp-sass) changed the part of its API that sets the Sass compiler to use. This changed the code in `gulpfile.js` to work with the new API.

[Gulp-sass](https://github.com/dlmanning/gulp-sass) also advised that [the node-fibers module*, used by our Sass compiler**, will not work past NodeJS v16](https://sass-lang.com/blog/node-fibers-discontinued). Removing it adds about 140 milliseconds to our build, on my machine at least, so I figured we should drop it now to save problems in the future.

This pins the `sass` module to v1.32.13 because versions after that produce warnings from the GOVUK Frontend Sass code that it looks like won't be changed for a while (see https://github.com/alphagov/govuk-frontend/issues/2238).

All these local changes are documented in the commits.

*The node-fibers module's official name is `fiber`.
**We use the Sass compiler recommended by https://sass-lang.com, Dart-sass. It's official name is `sass`.


benthorner (approved):
This looks OK to me:

```
make bootstrap
make test

# check how it looks
make run
```

The GOV.UK Frontend changes don't seem to affect us / are opt-in.

benthorner:
Forgot to mention: great explanations!

The only thing that's missing from the first commit is the part about pinning `sass`, which I didn't spot until I came back to the PR description. Putting it in the commit should make it easier for someone to reason about whether to upgrade it in the future, since the explanation of why it's on the version should be directly on the diff.

tombye:
> The only thing that's missing from the first commit is the part about pinning `sass`, which I didn't spot until I came back to the PR description. Putting it in the commit should make it easier for someone to reason about whether to upgrade it in the future, since the explanation of why it's on the version should be directly on the diff.

That's a great catch. I've added it to the commit that describes all the other fixes.

PR #146, opened 2021-07-23
Fix and standardise filtering on alert types

benthorner:

This is part of the work we need to do to replace (most of) the data.yaml
file with an API. Please see the commit messages for more details.

One outstanding question is whether we should ever show messages on the test
channel. Currently we don't have any, but if we did they would appear like any
other alert. We don't need to answer this question here, and I've made a note
on the card to make a decision as part of writing the API.

### DIFF
--- a/lib/alert.py
+++ b/lib/alert.py
@@ -35,10 +35,6 @@ class Alert(SerialisedModel):
         return self.is_current and self.is_public
 
     @property
-    def is_expired_or_test(self):
-        return self.is_expired or not self.is_public
-
-    @property
     def is_public(self):
         return self.channel in ['government', 'severe']
 
--- a/lib/alert.py
+++ b/lib/alert.py
@@ -9,7 +9,7 @@ from lib.alert_date import AlertDate
 class Alert(SerialisedModel):
     ALLOWED_PROPERTIES = {
         'identifier',
-        'message_type',
+        'channel',
         'starts',
         'sent',
         'expires',
@@ -40,7 +40,7 @@ class Alert(SerialisedModel):
 
     @property
     def is_public(self):
-        return self.message_type == 'alert'
+        return self.channel in ['government', 'severe']
 
     @property
     def is_expired(self):
--- a/lib/alert.py
+++ b/lib/alert.py
@@ -31,9 +31,19 @@ class Alert(SerialisedModel):
         return AlertDate(self.expires)
 
     @property
+    def is_current_and_public(self):
+        return self.is_current and self.is_public
+
+    @property
+    def is_expired_or_test(self):
+        return self.is_expired or not self.is_public
+
+    @property
+    def is_public(self):
+        return self.message_type == 'alert'
+
+    @property
     def is_expired(self):
-        if self.message_type == 'operator':
-            return True
         now = datetime.now(pytz.utc)
         return self.expires_date.as_utc_datetime < now
 
--- a/lib/alerts.py
+++ b/lib/alerts.py
@@ -13,8 +13,8 @@ class Alerts(SerialisedModelCollection):
         return [alert for alert in self if alert.is_current_and_public]
 
     @property
-    def expired_or_test(self):
-        return [alert for alert in self if alert.is_expired_or_test]
+    def expired(self):
+        return [alert for alert in self if alert.is_expired]
 
     @property
     def public(self):
--- a/lib/alerts.py
+++ b/lib/alerts.py
@@ -9,16 +9,16 @@ class Alerts(SerialisedModelCollection):
     model = Alert
 
     @property
-    def current(self):
-        return [alert for alert in self if alert.is_current]
+    def current_and_public(self):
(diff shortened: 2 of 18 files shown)


templates/components/alert.html
```diff
@@ -6,22 +6,22 @@
       <div class="alerts-icon__container alerts-icon__container--48">
         {{ alerts_icon(height=48, alert_active=alert.is_current) }}
         <h2 class="alerts-alert__title govuk-heading-s govuk-!-margin-bottom-3">
-          {% if alert.channel == 'operator' %}
-            Mobile network operator test
-          {% else %}
+          {% if alert.is_public %}
             <span class="govuk-visually-hidden">Emergency alert sent to </span>{{ alert.area_names | formatted_list(before_each='', after_each='') }}
+          {% else %}
+            Mobile network operator test
```
idavidmcdonald, line 12: I'm cautious that for an alert sent on the `test` channel, that we describe it as a Mobile network operator test.

Example messages that might be sent on the test channel could be a monthly test message to a small area of the UK (it might change which area got it). 

The two risks are
- the mobile network operators aren't happy with us describing them as the ones testing when it is actually us
- people who receive a message on the test channel are confused. For example, the page on mobile network operator tests says they are only for android phone users when technically an iphone user can opt in to messages on the test channel. 

I think this is tolerable for the moment though. When we send our next test channel message (whenever that might be), hopefully we will be looking at gov.uk/alerts and will spot if things look strange and can change it for the next one.

lib/alerts.py
```diff
@@ -13,8 +13,8 @@ def current_public(self):
         return [alert for alert in self if alert.is_current_public]
 
     @property
-    def expired_or_test(self):
-        return [alert for alert in self if alert.is_expired_or_test]
+    def expired(self):
```
idavidmcdonald, line 16: So if an operator channel message goes out, it will prompt people to go look at gov.uk/alerts. And when they get there, they will see nothing in current alerts and nothing in past alerts. At that point, what should they think? That the message they received is genuine?

I guess I just want to check that is definitely the decision we want?

idavidmcdonald (commented):
Good commit messages and some good additional test coverage. This is a good step in a good direction for gov.uk/alerts

I've left a few comments to be looked at please

lib/alert.py
```diff
@@ -30,10 +30,20 @@ def sent_date(self):
     def expires_date(self):
         return AlertDate(self.expires)
 
+    @property
+    def is_current_public(self):
+        return self.is_current and self.is_public
+
+    @property
+    def is_expired_or_test(self):
+        return self.is_expired or not self.is_public
```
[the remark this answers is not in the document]
benthorner: That's not quite true: an alert can also (potentially) be not started. We lose this property in a later commit anyway, but in principle we should be as explicit as we can to avoid future subtleties.

templates/components/alert.html
benthorner: Thanks for elaborating on the risks. I did agree this with @quis and as you say, the risk is low for now. My main concern was doing something to display the alerts properly if we get them: without this change they'll get shown like severe / government alerts, but without their own page (since they're not "public").

lib/alerts.py
benthorner: I agreed this with @quis, but now I'm not so sure. At the moment it makes sense because we manually change the site to say there are "planned tests". In future we don't want to be doing with that, so you're right there won't be any kind of confirmation about the alert until it expires. On the other hand, we don't want to alert people by saying "1 current alert" all in red. A complication is that these alerts don't have their own pages at the moment.

How about this (to avoid blocking this PR):

- We keep the change as-is (I'd argue it's a net improvement over treating all planned tests as "past alerts").

- We have a separate card to explore how to show planned tests when they happen. I'm thinking "1 current test", and letting the current-alerts page show them, potentially with more details e.g. the test area.

lib/alert.py
```diff
@@ -30,10 +30,20 @@ def sent_date(self):
     def expires_date(self):
         return AlertDate(self.expires)
 
+    @property
+    def is_current_public(self):
```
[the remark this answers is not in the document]
benthorner: Sure, I've amended with that 👍.

lib/alerts.py
idavidmcdonald: I'm happy to go with your suggestion if @quis is

idavidmcdonald (approved):
Thanks Ben!

lib/alerts.py
quis: Yup, this is good for now. We can still pre-announce planned tests manually for the moment, but at least they will go onto the past alerts page automatically with this strategy.

benthorner: OK, I'll merge this in.

Just to confirm, is it worth me writing a card to revisit this (as per my second bullet point)?

quis: Not yet. I don’t think we know enough about the frequency/nature of future testing.

PR #150, opened 2021-07-30
Resolve minor discrepancies in data.yaml vs API

benthorner:

Please see the commits for more details. We may want to have a discussion about the
removal of static map images before this goes in.


lib/alert.py
```diff
@@ -28,11 +29,15 @@ def approved_at_date(self):
...
+        return self.cancelled_at_date or self.finishes_at_date
+
+    @property
+    def finishes_at_date(self):
+        return AlertDate(self.finishes_at)
 
     @property
     def cancelled_at_date(self):
-        return AlertDate(self.cancelled_at)
+        return self.cancelled_at and AlertDate(self.cancelled_at)
```
quis, line 39: I’d prefer this return `None` rather than `False` if there isn’t a value present.

lib/alerts.py
```diff
@@ -8,6 +8,15 @@
 class Alerts(SerialisedModelCollection):
     model = Alert
 
+    def __init__(self, items):
+        sorted_items = sorted(
+            items,
+            reverse=True,
+            key=lambda item: item['starts_at']
```
quis, line 15: If you implement `__lt__` on the `Alert` class then you make the sorting reusable and don’t need the lambda or `reverse` here.

See this example from the admin app: https://github.com/alphagov/notifications-admin/blob/0363181daee439fe742fab6e1584d6a248b8df85/app/models/broadcast_message.py#L49-L62

quis:
Do you think it’s worth adding the logic to keep showing the static map if it’s in the data? When we move to the data coming from the API it will be missing, but:
- that’s still less time we’re making the map unavailable
- there’s always a chance we’ll have the dynamic map by then

lib/alert.py
benthorner: I agree that's more accurate. I've amended that commit to return `None` (by default).

lib/alerts.py
benthorner: We'd still want `reverse`, as we specifically want to show the list in date descending order. Doing it at the point of presentation is more explicit, though. I've rewritten the final commit to do that.

How does it look now?

benthorner:
> Do you think it’s worth adding the logic to keep showing the static map if it’s in the data?

This is what I was wondering. A downside is that it adds a bit of confusion for the person writing the API (do they need to specify the field at all?). I was thinking about the user need: we primarily need maps for current alerts, so a user can validate whether it applies to them. The value for historical alerts is less clear, and even for current alerts the fact that we're prepared to go public / dynamic without any maps suggests it's not worth the effort to keep these two.

I'd prefer to get rid of them unless we know someone is using them.

lib/alerts.py
quis: Moving it into the Jinja is nice 👍🏻

quis (approved):
If it’s quicker and less future unpicking to just remove the map images for now then let’s do that.

Thanks for the other amends.

PR #153, opened 2021-08-02
Filter alerts to ensure they're in the UK

benthorner:

While I'd like this to go in, we could also just use it for reference if we
think there's value in delaying the implementation.

### DIFF
--- a/data.yaml
+++ b/data.yaml
@@ -7,7 +7,7 @@ alerts:
     finishes_at: 2021-07-28T17:04:38+01:00
     content: |
       This is a mobile network operator test of the Emergency Alerts service. You do not need to take any action. To find out more, search for gov.uk/alerts
-    area_names:
+    general_area_names:
   - identifier:
     channel: operator
     approved_at: 2021-07-16T13:00:27+01:00
@@ -16,7 +16,7 @@ alerts:
     finishes_at: 2021-07-16T17:00:27+01:00
     content: |
       This is a mobile network operator test of the Emergency Alerts service. You do not need to take any action. To find out more, search for gov.uk/alerts
-    area_names:
+    general_area_names:
   - identifier:
     channel: operator
     approved_at: 2021-07-14T13:00:18+01:00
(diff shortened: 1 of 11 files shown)


quis:
I was thinking we could say `is_in_the_uk = len(general_areas) > 0`.

If we use https://github.com/alphagov/notifications-admin/pull/3980 then it will always return an empty list for alerts that don’t cover any part of the UK land area.

benthorner:
> it will always return an empty list for alerts that don’t cover any part of the UK land area

Hmm...it's the link between the UK and general areas seems quite subtle, so I'm hesitant to rely on it. For example, if a custom broadcast area fell within one of the  Santa Claus library areas, would it have ancestors (and thus general areas)?

It's still unclear what the data might look like for alerts created via the API, so I'd prefer not to make indirect decisions from it as that locks us in to certain implementations (and also as a general principle). Some other options:

- Get the Admin app to emit an explicit "is_in_uk" signal (more work, but less subtle).
- Augment area libraries to flag if the areas are non-public (and emit this with alerts).
- Just do the simple check here, maybe looking at all / more of the coordinates.

I lean towards the simple check, because it keeps our options open for the API. 

Do you think we need something more robust?

quis:
The overlapping code only looks at UK electoral wards, so it would never say an alert is in the general area of Santa Claus Village.

I guess this works fine for now. We might want to make it a bit more robust when we start letting people define their own shapes in the admin app, because the definition of ‘in the UK’ here is pretty broad.

quis (approved):
Good to go once the merge conflicts are sorted

benthorner:
> The overlapping code only looks at UK electoral wards

Are you sure? If I change [this line](https://github.com/alphagov/notifications-admin/pull/3980/files#diff-ef0345c352520fa964a7ab812d3489cc930473ee45b0123272f8ae96c1a98e16R169) to `test-santa-claus-village-rovaniemi-a` then Santa pops out in the result. I can see the code is different for custom areas, but again it's not impossible for a future custom area to overlap with a Santa one (although I haven't explored if the single-level hierarchy is important there).

Thanks for approving anyway. I agree this will tie us over for dynamic alerts, and for ones submitted via the API we could use your PR as a first / another line of defence to ensure the polygons are bounded by the UK.

PR #157, opened 2021-08-12
Add explainer film to the homepage

quis:
We’ve agreed with the comms folk that this video should go in the empty space on the homepage.

I’ve borrowed the responsive embed CSS from the admin app to make sure the video displays properly on different device sizes.

Hosted on Youtube, but served via their cookie-free domain so we don’t need to add a cookie policy.

image

image

***

https://www.youtube.com/watch?v=9yLd2AjGzYI

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -102,6 +102,13 @@
         <a href="/alerts/how-alerts-work" class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">How emergency alerts work</a>
       </p>
     </div>
+    <div class="govuk-grid-column-one-half">
+      <div class="responsive-embed responsive-embed--16by9">
+        <div class="responsive-embed__wrapper">
+          <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/9yLd2AjGzYI" frameborder="0" allowfullscreen></iframe>
+        </div>
+      </div>
+    </div>
   </div>
 {% endblock %}
 
--- a/src/index.html
+++ b/src/index.html
@@ -104,7 +104,7 @@
     </div>
     <div class="govuk-grid-column-one-half">
       <div class="responsive-embed responsive-embed--16by9">
-        <div class="responsive-embed__wrapper">
+        <div class="responsive-embed__content">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -102,6 +102,13 @@ <h2 class="govuk-heading-l">How it works</h2>
         <a href="/alerts/how-alerts-work" class="govuk-link govuk-link--no-visited-state govuk-!-font-weight-bold">How emergency alerts work</a>
       </p>
     </div>
+    <div class="govuk-grid-column-one-half">
+      <div class="responsive-embed responsive-embed--16by9">
+        <div class="responsive-embed__wrapper">
```
CrystalPea, line 107: I am probably missing something, but why is `<div class="responsive-embed__wrapper">` class div inside `<div class="responsive-embed responsive-embed--16by9">`? Shouldn't the wrapper be the outside one?

quis: This is an example of the [block-element-modifier (BEM)](https://css-tricks.com/bem-101/) methodology for writing CSS:

- The block `responsive-embed` is the top-level thing
- The element `responsive-embed__wrapper` is a child of the block (denoted by two underscores)
- The modifier `responsive-embed--16by9` alters the block in some way (denoted by two dashes)

Maybe `wrapper` isn’t the best name, but it sort of makes sense if you think of it as wrapping the `<iframe>` itself. What do you think?

src/index.html
```diff
@@ -102,6 +102,13 @@ <h2 class="govuk-heading-l">How it works</h2>
...
+          <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/9yLd2AjGzYI" frameborder="0" allowfullscreen></iframe>
```
rjbaker, line 108: Didn't know about the nocookie domain 👍

src/index.html
quis: I like `__outer` – changed in e1f0e13

rjbaker (approved):

src/index.html
CrystalPea: I like it, too, sorry for late reply 👍🏼

PR #158, opened 2021-08-12
Fix bugs in test_all_pages.py

tombye:
This work comes from @benthorner noticing a few bugs with our tests and a discussion me and him had following that about what they were and the best way to tackle them. In other words, it's not in pivotal :)

The bugs were (from the first commit comment):
In `test_local_links_lead_to_existing_routes_in_pages_with_alerts`, the alerts data instance was being passed to the ninja
environment wrapped in a list, which doesn't happen in the build.

In `test_all_pages_with_details_in_have_the_js_for_it`, the data used to make the alert added current or
past mutated the alert in a loop so all of it was applied at the same time.

We also decided that a test file named `test_all_pages.py` was generic enough that it encouraged the grouping of any miscellaneous tests, rather than developers choosing appropriate names/files for them. Because of that, this also splits `test_all_pages.py` into separate files, each named to identify the contents more specifically.

## Next steps

The assurances `test_local_links.py` gives should really just be handled by having a Jinja helper function that handles all URLs in pages. It could then include checking any local pages exist and that they are named correctly. This would be used like Flask's [url_for](https://flask.palletsprojects.com/en/2.0.x/api/#flask.url_for) is used in our admin app.

This is out of scope for this pull request so mentioning it here as a suggestion of possible next steps.


tests/test_details_component_has_js.py
```diff
@@ -0,0 +1,11 @@
+from glob import glob
```
benthorner, line 1: Could we move these tests under `tests/src`? This would be consistent with the other tests in that directory, and avoid cluttering the top level.

benthorner (approved):
Thanks for raising this @tombye ⭐.

Merging this would help with a PR I'm working on. I've rebased it against `main` and marked it as ready for review.

tests/test_details_component_has_js.py
benthorner: Fixed in https://github.com/alphagov/notifications-govuk-alerts/pull/163/commits/d9103f678c4511538055eb91c8832b43c23f5e37 and https://github.com/alphagov/notifications-govuk-alerts/pull/163/commits/fa45e96a122aab3212b6e34ab1053b5c71f39c26

PR #162, opened 2021-08-16
Fix Celery worker crashing after deployment

benthorner:

Resolves: https://github.com/alphagov/notifications-govuk-alerts/pull/151#discussion_r688341788

This is firstly because the startup command referred to a package
('main') that didn't exist, which we also then tried to import.


benthorner:
Tested in Staging. 

I can publish a task on the queue and it's processed as expected ✅ .

rjbaker (approved):

PR #163, opened 2021-08-17
More refactors and fixes to Celery app

benthorner:

This is the last batch. One thing I would still like to fix is the
random 'src' and 'templates' directories, together with the relative
paths e.g. app.utils.REPO, which don't work everywhere. That would
make this PR to big, so I'll make it a stretch goal.


app/config.py
```diff
@@ -42,3 +42,24 @@ class Config():
     STATSD_HOST = os.getenv('STATSD_HOST')
     STATSD_PORT = 8125
     STATSD_ENABLED = bool(STATSD_HOST)
+
+
+class Test(Config):
```
klssmith, line 47: Splitting up config by environment doesn't seem particularly necessary when only the test environment has different values. Unless there are plans to add different settings for development, preview or staging soon I wonder if it would keep things simpler to just add the environment variables here to `pytest.ini`. That way they can just be picked up by the `Config` class. The `FASTLY_SURROGATE_KEY`, which isn't an environment variable, could have the same value as the one in `Config`.

tests/conftest.py
```diff
@@ -40,11 +32,18 @@ def govuk_alerts():
...
     ctx.pop()
 
 
-def render_template(env, template_path, template_vars={}):
-    template = env.get_template(template_path)
-    content = template.render(template_vars)
-    return BeautifulSoup(content, 'html.parser')
+@pytest.fixture()
+def client_get(govuk_alerts, mocker):
+    mocker.patch('app.models.alerts.Alerts.load', return_value=Alerts([]))
```
klssmith, line 40: Optional - It might make it slightly more obvious what is going on and what is being mocked if this was in a different fixture, particularly since it gets overwritten in one of the tests.

This would mean that most tests would now need a fixture like `mock_no_alerts_returned` as well as the `client_get` fixture.

app/celery/tasks.py
```diff
@@ -1,21 +1,19 @@
...
 
-        upload_to_s3(current_app.config, rendered_pages)
-
-        purge_cache(current_app.config)
-    except Exception:
-        current_app.logger.exception("Failed to publish content to gov.uk/alerts: " + Exception)
+        upload_to_s3(rendered_pages)
+        purge_cache()
+    except Exception as e:
+        current_app.logger.exception("Failed to publish content to gov.uk/alerts: " + str(e))
```
klssmith, line 18: Could we get away with just `current_app.logger.exception("Failed to publish content to gov.uk/alerts")` here? `logger.exception` should give the stack trace so I'm not sure if adding `str(e)` to the message gives anything extra.

app/models/alerts.py
```diff
@@ -30,7 +30,11 @@ def last_updated_date(self):
         return AlertDate(self.last_updated)
 
     @classmethod
-    def from_yaml(cls, path):
+    def load(cls):
```
klssmith, line 33: It was unclear to me reading through this why this method exists when it solely calls another method. I can also see a mix of `Alerts.load()` and `Alerts.from_yaml()` in the code - I'm not sure how you've decided which one to use in each case. I understand this will change in future to have a different function, but it might be helpful to add a comment in the meantime.

tests/conftest.py
benthorner: Yeah, I was in two minds about this. I didn't seem worth cluttering most of the tests with the extra mock. 

A couple of other options:

- Add a comment above this line.
- Make it a fixture but require it here (still).

My thinking behind putting the mock here was based on providing a clean base for most tests to operate with. It's easier to do this in things like RSpec with a `before(:each)`. We could also do this:

```
@pytest.fixture(autouse=True)
def mock_no_alerts(): ...
```

But I'm a bit allergic to autouse fixtures because the tag is quite subtle.

Do you still lean towards a mock in each test?

app/celery/tasks.py
benthorner: Ah yes, and the stacktrace includes the `str()` of the exception 👍.

Updated in https://github.com/alphagov/notifications-govuk-alerts/pull/163/commits/f1173fbf5ee38157819cb0282996265bccde9a29.

app/models/alerts.py
benthorner: Sure, I've amended https://github.com/alphagov/notifications-govuk-alerts/pull/163/commits/354de0209831c5e242d90fa67ad1f76ca13b7862 to add a comment here 👍.

app/config.py
benthorner: Hmm...I'm not a fan of `pytest.ini` as I find it surprising to find config there. Setting the environment is fair enough. I was tempted to put config in `pytest.ini`, but this seemed more consistent with our other apps, but looking at Admin and API we seem to have a weird combination of both for no apparent reason 🤷.

I agree we can just hard-code the surrogate key. We'd need to read `DEBUG` from the environment, but that's not a big deal. I'm happy to switch to `pytest.ini` if that's more conventional.

Do you think it might be just me who gets confused by it?

tests/conftest.py
klssmith: I also don't really like `autouse`, so let's avoid that. At the moment, there's only one test which overwrites the mock and returns some data instead. How about leaving things as they are for now, but potentially splitting this out into a separate mock for clarity if we get to a situation where there are more tests which need `'app.models.alerts.Alerts.load'` to actually return some alerts.

app/config.py
klssmith: I'm not sure if it's more conventional or not. I think it would be possible to do away without config classes here in a way that we couldn't in api and admin. They both contain a lot of config variables that don't get loaded from environment variables, so we couldn't just use a `pytest.ini` file for the test config there.

I think the way this is written is fine - although it would save adding a class, after taking another look I'm not sure it would much difference in terms of making things simpler.

klssmith (approved):

app/config.py
benthorner: It's true we do have _fewer_ static / non-env variables, but we do still have variables like `DEBUG`, which are set statically but do need to change per environment. We could read that from the environment, but that creates its own smell: we're only ready it for the benefit of tests - it's not part of the production environment variable set.

I agree it's not clear cut either way 👍  

We can always change it in future if this approach turns out to be confusing.

PR #165, opened 2021-08-18
Add architecture doc

benthorner:

The diagram is the important part, while the text is more about the
history of the app (to help explain potential weirdnesses). In future
we may want to rewrite this to simply state the architecture and how
the app is organised, but the story format seems more appropriate as
we continue to make changes towards the architecture in the diagram.

Command to generate the diagrams [1]:

        cat arch.dot | docker run -i tsub/graph-easy

👉 [Rendered version](https://github.com/alphagov/notifications-govuk-alerts/blob/add-arch-doc-178801234/docs/architecture.md)

[1]: https://gist.github.com/benthorner/1901916c925e1b60cd14728c9aa5d332

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,0 +1,83 @@
+# Architecture
+
+⚠️  The following is about the _intended_ architecture of gov.uk/alerts, which may not (yet) reflect reality.
+
+The overall architecture looks like this:
+
+```
+                                     +-------------------------------+
+                                     |           data.yaml           |
+                                     |                               |
+                                     |    (legacy static alerts)     |
+                                     +-------------------------------+
+                                       |
+                                       |
+                                       v
++------------------------------+     +-------------------------------+     +--------------------------+
+|                              |     |                               |     |        Notify API        |
+| static pages (e.g. homepage) |     |     common rendering code     |     |                          |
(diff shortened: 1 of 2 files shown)


docs/architecture.md
```diff
@@ -0,0 +1,83 @@
...
+|                              |     |           S3 bucket           |
+|   (uploaded by Concourse)    | --> |                               |
++------------------------------+     +-------------------------------+
+                                       |
+                                       |
+                                       v
+                                     +-------------------------------+
+                                     |          CloudFront           |
+                                     |                               |
+                                     |   (tweaks caching headers)    |
```
rjbaker, line 43: Should we note that this is performed by Lambda@Edge response function?

rjbaker (approved):
Nice clear document 👍

docs/architecture.md
benthorner: I'd prefer not to here, as this is mostly to provide context about this repo. Mentioning that it "tweaks caching headers" should be enough of a prompt for someone to go looking the Terraform if they're interested.

We could always add a README to the Terraform module so it explains itself in more detail. Incidentally, I just found out about `terraform graph`, but I'm not sure anyone would ever want to see this 😬 :

test

PR #167, opened 2021-08-20
Simplify aggregate area names to be part of areas

benthorner:
https://www.pivotaltracker.com/story/show/178986763

Previously this was a separate list due to some confusion about the
"areas" field being a list (of lists). In reality it's just a dict
with a "simple_polygons" field, so we can co-locate aggregate area
names inside it as another field.


rjbaker (approved):
Intent is definitely clearer 👍

PR #169, opened 2021-08-24
Move HTML file to more conventional locations

benthorner:

This is the final PR to finish converting this app to something
that better ressembles our other Flask / Celery apps. It still
works a bit differently, but having files in the usual places
will reduce the learning curve for future devs.

"dist" is the one directory that still sticks out: conventionally
it should be "app/static". Changing that would also mean tweaking
the Concourse pipeline [1]. Leaving it as-is doesn't seem like it
will cause any major confusion.

[1]: https://github.com/alphagov/notifications-broadcasts-infra/blob/7758409ff1a4548a2507e0f26f87f7e2afbf5fd3/ci/govuk-alerts.yml#L202


rjbaker (approved):
Good tidy up! I guess this was always a necessary step as the "app" has transitioned from a static site builder to an odd hybrid. i.e. the `dist` directory. Makes sense to keep it consistent with everything else as far as reasonably possible.

PR #173, opened 2021-09-01
Improve grouping of past alerts

quis:
From this:

image

***

To this (grouping the alerts by date heading):

image

***

To this (only showing one operator test per day):

image

### DIFF
--- a/app/models/alert_date.py
+++ b/app/models/alert_date.py
@@ -39,3 +39,7 @@ class AlertDate(object):
     @property
     def as_local_datetime(self):
         return self._local_datetime
+
+    @property
+    def as_local_date(self):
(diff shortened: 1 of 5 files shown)


rjbaker:
Should we display the time with each alert? Can be useful in cases where we've sent multiple.

quis:
@rjbaker Because this combines all the operator tests for a day into one it’s not possible to give the time.

The most useful thing to communicate about these operator tests is that, on a given day, there was a test and that if you received something it was expected.

Our experience is that it confuses more than it helps to be very specific about these tests, for example saying where/if they are geographically targeted.

rjbaker (approved):

PR #175, opened 2021-09-06
Automate planned tests page

quis:
At the moment, whenever there is a planned test we need to update and deploy gov.uk/alerts manually.

We also update the site manually to announce upcoming public tests, like Reading.

This commit:
- automatically puts a banner on the homepage and updates the ‘Planned tests’ page whenever there is an operator test happening on a given day
- gives us a way of publishing details of upcoming public tests without having to directly edit HTML (it’s yet another YAML file)

I’ve even made it work in the case when there’s an upcoming public test, and a current operator test, like we had in the run-up to Reading.

For operator tests the banner will stay up until the end of the day. I think this is sensible since the test might not last for very long, and there might be several in one day (which we treat as one). For this to work in a completely automated way we will need to schedule republishing of the site at 00:01 each day.

If we do this then the MNOs don’t need to arrange their planned tests with us in advance, reducing our support burden.

With some fake data:

image

image

image

***

<sup>Sorry for the mega commit</sup>

### DIFF
--- a/app/models/alert_date.py
+++ b/app/models/alert_date.py
@@ -1,4 +1,10 @@
-from notifications_utils.timezones import utc_string_to_aware_gmt_datetime
+import datetime
+
+import pytz
+from notifications_utils.timezones import (
+    local_timezone,
+    utc_string_to_aware_gmt_datetime,
+)
 
 
 class AlertDate(object):
@@ -7,6 +13,18 @@ class AlertDate(object):
     def __init__(self, _datetime):
         self._datetime = _datetime  # a timezone-aware UTC datetime
         self._local_datetime = utc_string_to_aware_gmt_datetime(self._datetime)
+
+    def __eq__(self, other):
+        return self._datetime == other._datetime
+
+    def __lt__(self, other):
+        return self._datetime < other._datetime
+
+    def __hash__(self):
+        return hash(self._datetime)
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}({self._datetime})'
 
     @property
     def time_as_lang(self):
@@ -44,6 +62,10 @@ class AlertDate(object):
         return self._local_datetime.isoformat()
 
     @property
+    def as_iso8601_date(self):
+        return self._local_datetime.date().isoformat()
+
+    @property
     def as_utc_datetime(self):
         return self._datetime
 
@@ -54,3 +76,21 @@ class AlertDate(object):
     @property
     def as_local_date(self):
         return self._local_datetime.date()
+
+    @property
+    def at_midday(self):
+        return self.__class__(
+            datetime.datetime.combine(
+                self.as_local_date,
+                datetime.time(hour=12, minute=0),
+                tzinfo=local_timezone,
+            ).astimezone(pytz.utc)
+        )
+
+    @classmethod
+    def now(cls):
+        return cls(datetime.datetime.utcnow())
+
+    @property
+    def is_today(self):
+        return self.as_local_date == self.now().as_local_date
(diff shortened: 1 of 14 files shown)


app/models/alert_date.py
```diff
@@ -8,6 +14,18 @@ def __init__(self, _datetime):
         self._datetime = _datetime  # a timezone-aware UTC datetime
         self._local_datetime = utc_string_to_aware_gmt_datetime(self._datetime)
 
+    def __eq__(self, other):
+        return self._datetime == other._datetime
+
+    def __lt__(self, other):
```
CrystalPea, line 20: is there a reason for these function names to be abbreviated? I'm asking as it makes them less readable.

Also I can't seem to see where we use them 😳

CrystalPea (commented):
Good work towards automating this, well done! 

It's not fully automated yet though, right? Since we will have to update the planned alerts yaml file manually? 

Do we plan, in the future, to add some kind of option to Notify to allow operators to schedule alerts, and gov.uk/alerts will poll API for those, like we do with alerts that are already happening?

Also this is a lot of code and tests in one commit, that makes it harder to read. In the future it would be cool to have several smaller commits, each with part of the work and associated tests, then it would be easier to review this, and probably commit messages could have more technical details then, too.

Oh yeah, I left a comment about some internal methods, not sure where we use them (or are we modifying some methods that are only used under the hood?)

app/models/alert_date.py
quis: `__eq__` and `__lt__` are examples of ‘dunder’ (**d**ouble **under**score) methods which have a special meaning in Python. 

`__eq__` is called on an object when you use the equality operator, for example these two lines are the same:
```
AlertDate(datetime.utcnow()) == 1
AlertDate(datetime.utcnow()).__eq__(1)
```

Without defining `__eq__`:
```
now = datetime.utcnow()
AlertDate(now) == AlertDate(now)
> False
```

With `__eq__` defined:
```
now = datetime.utcnow()
AlertDate(now) == AlertDate(now)
> True
```

***

`__lt__` is called on an object when you use the less than operator, for example these two lines are the same:
```
AlertDate(datetime.utcnow()) < 1
AlertDate(datetime.utcnow()).__lt__(1)
```

***

Implementing both `__eq__` and `__lt__` makes an object _sortable_. Which means you can do this:
```
sorted([
    AlertDate(datetime.utcnow() + timedelta(hours=2)),
    AlertDate(datetime.utcnow() + timedelta(hours=1)),
    AlertDate(datetime.utcnow() + timedelta(hours=7)),
])
> [
    AlertDate(2021-11-03 11:58:41.288081),
    AlertDate(2021-11-03 12:58:41.288030),
    AlertDate(2021-11-03 17:58:41.288099),
]
```

Without `__eq__` and `__lt__` implemented you get this:
```
sorted([
    AlertDate(datetime.utcnow() + timedelta(hours=2)),
    AlertDate(datetime.utcnow() + timedelta(hours=1)),
    AlertDate(datetime.utcnow() + timedelta(hours=7)),
])
> TypeError: '<' not supported between instances of 'AlertDate' and 'AlertDate'
```

So we don’t call these methods directly, but the [algorithm of the `sorted` method](https://en.wikipedia.org/wiki/Timsort) does.

quis:
@CrystalPea when operators are doing tests we are not going to announce these in advance. They will just appear as a ‘planned test’ at the same time as they happen, then stay on the site for the rest of the day.

We’ll only use the YAML file for announcing tests on public channels in advance.

We probably need some stuff in the runbook for this.

app/models/alert_date.py
CrystalPea: Ahh, makes sense, thanks so much for the explanation!

CrystalPea (approved):
Thanks for extra context, and I agree, it would be good to have something in the runbook about the planned tests.

PR #176, opened 2021-09-08
Add planned operator test for 9/9/21

idavidmcdonald:
Note, there will be two alerts sent, one at 10 and one at 10:15
but we can potentially just tell the user it will be 1 planned test.
Will need sign off from Karl.

### DIFF
--- a/app/templates/views/planned-tests.html
+++ b/app/templates/views/planned-tests.html
@@ -45,7 +45,7 @@
         Some mobile phone networks in the UK will test emergency alerts between 10am and 11am.
       </p>
       <p class="govuk-body">
-        Most phones and tablets will not get a test alert.
+        Most phones and tablets will not get a test alert. A small number of people may receive 2.
       </p>
       <p class="govuk-body">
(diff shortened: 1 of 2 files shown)


app/templates/views/planned-tests.html
```diff
@@ -38,12 +38,11 @@
...
       <h2 class="govuk-heading-m govuk-!-margin-top-6">
-        [day of week] [day of month] [month] [year]
+        Thursday 9 September 2021
       </h2>
       <p class="govuk-body">
-        Some mobile phone networks in the UK will test emergency alerts between 1pm and 2pm.
+        Some mobile phone networks in the UK will test emergency alerts between 10am and 11am.
       </p>
       <p class="govuk-body">
         Most phones and tablets will not get a test alert.
```
karlchillmaid, line 48:
```suggestion
        Most phones and tablets will not get a test alert. A small number of people may receive 2.
```

karlchillmaid (approved):

PR #177, opened 2021-09-08
Update launch window from 'summer' to 'autumn'

karlchillmaid:
The existing content is out of date.

We’ve been waiting for confirmation that we can change the expected launch window from 'summer' to 'autumn'. We got that confirmation today.


idavidmcdonald (approved):
Looks good

PR #178, opened 2021-09-09
Add planned test for 10 September and add alerts that were sent on 9 September

idavidmcdonald:
Review commit by commit

### DIFF
--- a/data.yaml
+++ b/data.yaml
@@ -1,6 +1,24 @@
 alerts:
   - identifier:
     channel: operator
+    approved_at: 2021-09-09T10:15:00+01:00
+    starts_at: 2021-09-09T10:15:00+01:00
+    cancelled_at: 2021-09-09T10:30:00+01:00
+    finishes_at: 2021-09-09T10:30:00+01:00
+    content: |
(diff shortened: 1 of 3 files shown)


data.yaml
```diff
@@ -1,4 +1,22 @@
 alerts:
+  - identifier:
+    channel: operator
+    approved_at: 2021-09-09T10:15:00+01:00
+    starts_at: 2021-09-09T10:15:00+01:00
+    cancelled_at: 2021-09-09T10:30:00+01:00
+    finishes_at: 2021-09-09T10:30:00+01:00
+    content: |
+      This is a mobile network operator test of the Emergency Alerts service. You do not need to take any action. To find out more, search for gov.uk/alerts
+    areas: {}
```
CrystalPea, line 10: is it intended that areas are empty?

idavidmcdonald: I copied all the other operator alerts below which are the same so I assume it is OK because we don't show which areas operator tests get sent to.

If it was on a severe channel then it would matter.

leohemsted (approved):

data.yaml
CrystalPea: Makes sense, thanks! 👍🏼

CrystalPea (approved):
LGTM

PR #180, opened 2021-09-14
Create entry for Suffolk severe alert with correct identifier 

CrystalPea:
the alert was broadcast on 25-may-2021, but for some reason our identifier for it was 3-may-2021. Now we are fixing the identifier and leaving an empty identifier for 3-may-2021.

If user goes to url ending alerts/3-may-2021 they should be redirected to alerts/25-may-2021.

This is the first part of that work - creating an entry for that alert with correct identifier while keeping the old entry.

We will deploy this just before deploying the lambda change to support redirect for that alert, and after that lambda change is deployed, we will deploy another PR to get rid of the old entry.

### DIFF
--- a/data.yaml
+++ b/data.yaml
@@ -178,6 +178,23 @@ alerts:
     areas:
       aggregate_names:
         - East Suffolk
+  - identifier: 25-may-2021
+    channel: severe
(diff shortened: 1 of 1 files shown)


benthorner (approved):
Nicely explained 👍

PR #181, opened 2021-09-15
 Remove old entry for severe Suffolk alert

CrystalPea:
If someone goes to the address for the old entry, we will redirect them.

Needs to go in after https://github.com/alphagov/notifications-govuk-alerts/pull/180 and after we deploy lambda changes to support redirects.

TODO: make a note in the README that if you want to put in a redirect, you do it through Terraform/lambda in the infra repo

### DIFF
--- a/docs/redirects.md
+++ b/docs/redirects.md
@@ -1,0 +1,15 @@
+# Redirects
+
+If you need to set up a new redirect, you can do that in our notifications-broadcasts-infra repository.
+
+In there, head to `terraform/modules/govuk-alerts-website/files/redirects` and add your redirect to REDIRECTS constant, using the following format:
+
+```
+REDIRECTS = {"<uri of page you want to redirect from>": "<uri of page you want to redirect to>"}
+```
+
+Deploy-wise, you will need to do 3 deploys in quick succession:
+
+1. deploy a PR that adds the page to redirect to
+2. deploy the lambda change (through govuk-alerts-infra Concourse pipeline)
+3. deploy a PR that removed the old page we want to get rid of
--- a/README.md
+++ b/README.md
@@ -57,5 +57,6 @@ To run a specific JavaScript test, you'll need to copy the full command from `pa
 
 ## Further documentation
 
-- [Image optimisation](docs/image-optimisation.md)
 - [Architecture](docs/architecture.md)
+- [Image optimisation](docs/image-optimisation.md)
+- [Setting up redirects](docs/redirects.md)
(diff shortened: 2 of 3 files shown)


docs/redirects.md
```diff
@@ -0,0 +1,15 @@
...
+
+If you need to set up a new redirect, you can do that in our notifications-broadcasts-infra repository.
+
+In there, head to `terraform/modules/govuk-alerts-website/files/redirects` and add your redirect to REDIRECTS constant, using the following format:
+
+```
+REDIRECTS = {"<uri of page you want to redirect from>": "<uri of page you want to redirect to>"}
+```
+
+Deploy-wise, you will need to do 3 deploys in quick succession:
```
leohemsted, line 11: "quick succession" scares me.

How quick? What happens if it's slow, because you get distracted, have wifi failure, functional tests are flaky, etc? Do we need to worry about fastly caching timelines in this?

docs/redirects.md
```diff
@@ -0,0 +1,15 @@
...
+
+1. deploy a PR that adds the page to redirect to
+2. deploy the lambda change (thorugh terraform)
```
leohemsted, line 14: minor typo
```suggestion
2. deploy the lambda change (through terraform)
```

also, do you need to know that it's through terraform? is it worth saying "through the govuk-alerts-infra deploy pipeline" or something?

docs/redirects.md
CrystalPea: well we have a double entry if we are stuck in the middle (old page and new page both exist) - alternative way would be to not have a redirect for a short while? Then we only need 2 PRs, not 3, and we never have a confusing-looking list, and it is unlikely someone will use saved up link in that time. What do you think?

docs/redirects.md
CrystalPea: Ahh, I think I have been only remotely aware of govuk-alerts-infra concourse pipeline. So it deploys terraform for us? Where can I find this pipeline? I don't see it with our other pipelines?

CrystalPea: cool, thanks for the link, I will update the guidance 👍🏼

docs/redirects.md
leohemsted: i think three deploys should stay the same. As a team we're used to doing multiple PRs for zero-downtimey-ness when dealing with API changes, celery tasks, etc. I don't think there are any downsides to things being in a mid-state for a little while while PRs are reviewed, tests run, etc.

Is there a risk to having duplicate pages with no redirects between them? That seems fine to me as long as the content still makes sense, but there's some things I don't know the answer to like: fastly caching, google search indexing robots scraping pages, etc. Thoughts?

I don't think anyone here expects things to take days and days, but when I see "quick" I guess I start worrying about co-ordinating concourse jobs, etc. Or should we panic if it was quick but then concourse crashed. etc. And I _think_ for this the answer is "no we shoudn't panic if it takes a bit longer or someone goes on lunch during the deploy etc"

CrystalPea: yeah, not quick as very time sensitive, quick as relatively quick - I added a bit more context to the guidance.

idavidmcdonald (approved):

PR #184, opened 2021-09-17
Take down emergency alert test from 17 sept.

idavidmcdonald:
Reverts alphagov/notifications-govuk-alerts#182 and adds the two tests to the list of past alerts data.yaml file.

### DIFF
--- a/data.yaml
+++ b/data.yaml
@@ -1,6 +1,24 @@
 alerts:
   - identifier:
     channel: operator
+    approved_at: 2021-09-17T11:00:00+01:00
+    starts_at: 2021-09-17T11:00:00+01:00
+    cancelled_at: 2021-09-17T11:15:00+01:00
+    finishes_at: 2021-09-17T11:15:00+01:00
+    content: |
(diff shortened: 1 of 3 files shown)


data.yaml
```diff
@@ -1,4 +1,22 @@
 alerts:
+  - identifier:
+    channel: operator
+    approved_at: 2021-09-17T11:00:00+01:00
+    starts_at: 2021-09-17T11:00:00+01:00
+    cancelled_at: 2021-09-17T11:15:00+01:00
+    finishes_at: 2021-09-17T11:15:00+01:00
```
benthorner, line 7: Optional: this would actually be the start time + the hard-coded max duration. We don't show this data anywhere, so it's not a big deal.

(Similarly with the alert below.)

benthorner (approved):

data.yaml
idavidmcdonald: Oh thanks for pointing out! I did not realise that. I won't make the change as in time these will be replaced with correct data from the API if used at all.

PR #187, opened 2021-09-22
Dynamic slugs

leohemsted:
Probs best viewed commit by commit.

text here to come tomorrow


app/render.py
```diff
@@ -17,6 +17,35 @@
...
+    Gets the url slug for an alert (given the global alerts object obtained from `Alerts.load()`.)
+
+    Note: The alert must be public to have a url slug.
+
+    Will return a date string in the style '3-jun-2021', unless there were already alerts that day, in which case it'll
+    append a 1-indexed counter on to the end eg '3-jun-2021-2'. Note that the first alert of the day never has a count.
+    """
+    alerts_on_this_day = sorted([
+        other_alert
+        for other_alert in alerts.public
```
leohemsted, line 38: Question (possibly for @quis / @karlchillmaid )

We're only assigning URLs for public alerts (severe/government channel).

If we ever want specific URLs for other alerts (operator/test/rmt/exercise etc), then we need to make that decision _before_ we send both a public and a test alert on the same day. (As the URLs will be affected because they're numbered). After that point we'll need to stick with this URL scheme. We could always have operator alerts be on a different naming scheme entirely, eg `/alerts/operator/25-may-2021`.

The decision doesn't need to happen right now, but thought it was worth flagging in case we ever want to have individual pages for operator tests.

app/models/alert_date.py
```diff
@@ -24,6 +24,18 @@ def date_as_lang(self):
...
         return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y}'
 
+    @property
+    def as_url(self):
+        """
+        * non-zero padded day
+        * lower case month abbrevation
+        * full year
+
+        3-jun-2021
```
quis, line 34: Should check if there’s precedent on GOV.UK for using `jun` or `june` in URLs

app/render.py
quis: Unlikely that’d we’d give non-public alerts a URL, but a prefix/suffix/different structure feels right.

app/models/alert_date.py
idavidmcdonald: @karlchillmaid please do confirm if a change is needed to the formatting of the month

idavidmcdonald: Just spotted that we have a url that goes against this convention in production
https://www.gov.uk/alerts/29-june-2021

Meaning either this PR needs to change to not use the month abbreviation or we need to change the url to `jun` and setup a redirect for that alert

karlchillmaid: From the document we wrote when we agreed the URL structure in July:

> The month will always be abbreviated to 3 letters.
> ...
> We discussed whether to format the month as a word (in keeping with GOV.UK style) or a number. We felt that a mixture of words and numbers would make the URL friendlier.
> ...
> We also discussed whether we should use the full month or a 3 letter abbreviation. For example ‘september’ or ‘sep’. The GOV.UK style guide says: ‘when space is an issue - in tables or publication titles, for example - you can use truncated months: Jan, Feb’.

Therefore it should be `jun` instead of `june` – it’s always tricky with the shorter months, but we should always consider the worst case (September) as our baseline.

app/render.py
karlchillmaid: Agree

rjbaker: RMT "test" alerts are public and should be published as they'll be received by real users.

app/models/alert_date.py
idavidmcdonald: Karl has confirmed we are going for abbreviation still

I have now deployed a redirect for https://www.gov.uk/alerts/29-june-2021 to https://www.gov.uk/alerts/29-jun-2021 so I think we can now move forward on this.

tests/app/main/views/test_alert.py
```diff
@@ -19,3 +25,53 @@ def test_alert_breadcrumbs(
...
+            content='test 1',
+            starts_at=datetime(2021, 4, 21, 11, 0, tzinfo=pytz.utc),
+            cancelled_at=None,
+            finishes_at=datetime(2021, 4, 21, 15, 0, tzinfo=pytz.utc),
+        )
+    ]))
+
+    html = client_get('alerts/21-apr-2021')
+    assert html.select_one('main h2').text.strip() == 'Stopped sending at 4:00pm on Wednesday 21 April 2021'
+    assert html.select_one('main h2').text.strip() == 'Stopped sending at 4:00pm on Wednesday 21 April 2021'
```
idavidmcdonald, line 60: Duplicate assertion here

idavidmcdonald (commented):
Generally looking great. Thanks for this.

Most of my comments are optional. There is one that I feel a bit stronger than optional on (but not so strong that I would never approve if it didn't change).

tests/app/models/test_alert_date.py
```diff
@@ -8,12 +8,13 @@
 
 
 def test_AlertDate_properties():
-    sample_datetime = datetime(2021, 3, 21, 10, 30, tzinfo=pytz.utc)
+    sample_datetime = datetime(2021, 3, 2, 10, 30, tzinfo=pytz.utc)
```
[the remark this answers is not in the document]
leohemsted: i wanted to test that the day was not zero-padded in `as_url`, but wasn't sure if it was worth creating an explicit test for. maybe it is, given that these tests are low footprint and quick to run? 🤔

app/models/alert_date.py
```diff
@@ -24,6 +24,18 @@ def date_as_lang(self):
         dt = self._local_datetime
         return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y}'
 
+    @property
+    def as_url(self):
```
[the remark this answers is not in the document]
leohemsted: get_url_for_alert is used to get a url from an alert - this is a property of the various timestamps within an alert. I don't think you should ever be using this in a context where you could confuse them.

That said, you shouldn't ever call this for a `cancelled_at_date` either, only a `starts_at_date`, so i'm not completely against moving this out of the `AlertDate` class and on to a property on the Alert object. It just felt neat on the AlertDate as it doesn't care about the rest of the alert object.

then it could be `Alert.start_date_for_url` or something like that perhaps? I think that my thought process when naming this function was that I'd like to keep reference to URL in the function definition just so it's obvious that it's got a bigger impact - i feel like we're generally aware that URLs are harder to change than other things. (Whereas a `short_date` for example someone may come along later and think "actually i prefer two digit year to four digit year" and cause irreversible damage).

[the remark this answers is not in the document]
idavidmcdonald: Ah I see that I had misunderstood. I thought the `as_url` was a property on the `Alert` rather than on the `AlertDate` class.

I am less worried now about what you currently have written and don't think it needs any changes.

Thanks for this.

app/render.py
```diff
@@ -17,6 +17,35 @@
...
+        if other_alert.starts_at_date.as_local_datetime.date() == alert.starts_at_date.as_local_datetime.date()
```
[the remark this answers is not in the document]
idavidmcdonald: This comment is slightly incorrect now I realised `as_url` is on the `AlertDate` class. It would need to be `other_alert.starts_at.as_url == alert.starts_at.as_url`

Whether that is particularly advantageous I'm not sure

data.yaml
```diff
@@ -98,7 +109,8 @@ alerts:
     content: |
       This is a mobile network operator test of the Emergency Alerts service. You do not need to take any action. To find out more, search for gov.uk/alerts
     areas: {}
-  - identifier:
+  - id: 4c419ff4-8c74-46d9-89f3-141e65fa2693
```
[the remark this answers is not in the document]
leohemsted: Correct. all these IDs were generated using `uuidgen | tr '[:upper:]' '[:lower:]' | pbcopy`

app/render.py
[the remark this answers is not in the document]
leohemsted: i think i'll stick to comparing the rich date class as that feels more correct. Though I will change it from `as_local_datetime.date()` to `as_local_date` as we've got the property available already

tests/app/test_render.py
```diff
@@ -0,0 +1,77 @@
...
+        get_url_for_alert(alerts[0], alerts)
+
+
+@pytest.mark.parametrize('index, expected_url', [
+    (0, '20-apr-2021'),
+    (1, '21-apr-2021'),
+    (2, '21-apr-2021-2'),
+    (3, '21-apr-2021-3'),
+    (4, '22-apr-2021'),
+])
```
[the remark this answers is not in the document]
leohemsted: interestingly, probably due to python reusing objects in memory through cache wizadry or something , the first test to run takes 0.16s, and then the rest of the parametrised tests all take <0.005s. so there wouldn't really be any time saving. (you can use `--durations=10` to show the 10 slowest tests)

not doing in a for loop has the advantage of testing all even if one fails, which is why i wrote it like this, so i could run and see when i made changes which ones broke and which ones fixed.

there are some other very slow tests in this suite which i think might render pages or load yaml or similar expensive operations in nested for loops (they're all `tests/app/main/test_local_links.py::test_local_links_lead...`) but this one doesn't really concern me and it made my debugging easier when writing at least

tests/app/models/test_alert_date.py
```diff
@@ -24,6 +25,7 @@ def test_AlertDate_properties_work_with_bst():
     assert alerts_date.as_iso8601 == '2021-04-21T11:30:00+01:00'
     assert alerts_date.as_utc_datetime == sample_datetime
     assert alerts_date.as_local_datetime == datetime_in_bst
+    assert alerts_date.as_url == '21-apr-2021'
```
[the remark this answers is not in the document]
leohemsted: good point, I updated the test to run at `2021-04-20T23:30` to cover this case 👍

idavidmcdonald (approved):

PR #195, opened 2021-10-11
Add dynamic datasource of alerts from Notify API

benthorner:
https://www.pivotaltracker.com/story/show/178801357

This replaces most of the alerts in data.yaml. I had to make a
change to the way we deal with timestamps to support this. This
PR also includes a couple of one-line fixes.

Please see the commits for more details.

- Tested and works with local API ✅.
- Tested on Preview (works with `flask publish`) ✅.


data.yaml
```diff
@@ -1,232 +1,34 @@
...
-      To find out more, call 0808 1697692 or search for gov.uk/alerts
-    areas:
-      aggregate_names:
-        - Reading
-### vvv DO NOT DELETE: NOT IN DB vvv ###
   - id: 52499a28-c5ca-492c-ba3c-d7ac820b06cf
     channel: operator
-    approved_at: 2021-06-22T09:00:00
-    starts_at: 2021-06-22T09:00:00
+    approved_at: 2021-06-22T09:00:00Z
```
rjbaker, line 19: Useful to know these require the offset or at least be specified as being UTC 👍

PR #197, opened 2021-10-15
Extend publish recipe to check it was successful

benthorner:
This repeats an approach we have to check API migrations have been
successful [1][2]. Rather than split the code across the Makefile
and a script in Concourse, I thought it was worth the effort to
make the existing recipe behave synchronously. This means we don't
need to make any changes to Concourse, but does mean the script is
a bit more compact than the one we have for API. Hopefully all the
comments are enough to make it maintainable in the future.

[1]: https://github.com/alphagov/notifications-api/blob/c84daf0b7bdb5b348475571bf5dd81c04770886a/Makefile#L159
[2]: https://github.com/alphagov/notifications-aws/blob/fc34c0493b60887d32d2aa9fef67dbaa18eb6e7f/concourse/templates/api.yml.j2#L150-L166

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,25 @@ cf-deploy: cf-target ## Deploys the app to Cloud Foundry
 
 cf-run-task-publish: cf-target
 	cf run-task notify-govuk-alerts -c 'flask publish'
+	# run a shell to setup a command to check progress;
+	# "eval" is hacky way of assigning some text we want
+	# to substitute later - by evaluating what follows as
+	# a mini makefile and merging its state with this one
+	$(eval \
+		CHECK_COMMAND := cf curl "/v3/apps/`cf app --guid notify-govuk-alerts`/tasks?order_by=-created_at" | jq -r ".resources[0].state" \
+	)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -76,6 +76,25 @@ cf-deploy: cf-target ## Deploys the app to Cloud Foundry
...
+	)
+	# run a little shell script to check if the task is
+	# successful and return a non-zero status if not;
+	# make doesn't support multiline scripts, so the "\"
+	# and ";" turn it into a oneliner with an exit status
+	for _ in 1 2 3 4 5; do \
+		if $(CHECK_COMMAND) | grep SUCCEEDED; then\
+			exit 0; \
+		fi; \
+		sleep 1; \
```
sakisv, line 94: Previous sleep period was [set to 10](https://github.com/alphagov/notifications-broadcasts-infra/pull/326/files#diff-ff09dc3530139d0bfd5a645dc5c7afb5d7710dd65cacbde64e6ddc2163bc0ebaL281). It was probably chosen arbitrarily, but I just wanted to point out the discrepancy here.

PR #199, opened 2021-11-04
Hide content and areas of operator tests

quis:
We want to always show a standard message for operator tests on the planned tests page, not the content of the alert itself.

We also don’t want to show the names of the areas targeted by an alert, because past experience has shown that people interpret this to mean that they are more likely to receive the test than they actually are.

### DIFF
--- a/app/models/alert.py
+++ b/app/models/alert.py
@@ -6,21 +6,12 @@ from notifications_utils.serialised_model import SerialisedModel
 from app.models.alert_date import AlertDate
 
 
-class BaseAlert(SerialisedModel):
+class Alert(SerialisedModel):
     ALLOWED_PROPERTIES = {
-        'starts_at',
-    }
-
-    @property
-    def starts_at_date(self):
-        return AlertDate(self.starts_at)
-
-
-class Alert(BaseAlert):
-    ALLOWED_PROPERTIES = BaseAlert.ALLOWED_PROPERTIES | {
         'id',
         'channel',
         'approved_at',
+        'starts_at',
         'cancelled_at',
         'finishes_at',
         'areas',
@@ -46,6 +37,10 @@ class Alert(BaseAlert):
     @property
     def approved_at_date(self):
         return AlertDate(self.approved_at)
+
+    @property
+    def starts_at_date(self):
+        return AlertDate(self.starts_at)
 
     @property
     def expires_date(self):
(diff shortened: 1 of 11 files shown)


app/models/alert.py
```diff
@@ -9,7 +9,6 @@
 class BaseAlert(SerialisedModel):
     ALLOWED_PROPERTIES = {
         'starts_at',
-        'content',
```
benthorner, line 12: How come the field is diverging between the two subclasses now? It would be simpler to use the same field name for both kinds of alert, especially since they're pretty much synonyms of each other - I'm not sure what the difference is.

quis: `content` is the message the user will get on their phone. `description` is information _about_ the alert, for example when it will happen and who will receive it. I’m not wedded to that naming though, or the particular class structure if you can think of a clearer way of expressing it.

For context here is where these things appear, and how they are shown differs from one place to another:

— | Real alert | Operator alert | Upcoming test 
----|-----------|----------------|-------------- 
**Instance** | `Alert` | `Alert` | `PlannedTest`
**‘Past alerts’ page** | `content` shown | `content` shown | Not shown
**‘Planned tests’ page** | Not shown | Generic text shown | `description` or generic text shown

benthorner: Aha, I see where you're coming from now. The field name makes sense with your explanation, but what's confusing it for me is the class hierarchy as this brings it into conflict with `content`.

Looking through the code, we don't seem to be using many of the normal alert fields for planned tests anyway. Arguably a planned test isn't an `Alert`, either: it may actually represent many alerts. 

I'm wondering how hard it would be to just make `PlannedTest` a child of `Object` and copy across the small amount of functionality it shares with `Alert`. Then we get more flexibility in future too.

What do you think?

quis: I’ve tried that in 16557aa and think it’s an improvement. Means we get to keep the pattern of having on class per file as well, which is nice.

app/templates/views/planned-tests.html
```diff
@@ -42,13 +42,13 @@ <h1 class="govuk-heading-xl">
       <h2 class="govuk-heading-m govuk-!-margin-top-6">
         {{ alert.starts_at_date.date_as_lang }}
       </h2>
-      {% if alert.content %}
+      {% if alert.description %}
```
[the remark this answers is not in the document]
quis: Yes, agree, and on further reflection `PlannedTests` should be `UpcomingPublicTests`, so the ‘Planned tests’ page shows a combination of operator `Alert`s from the API and `UpcomingPublicTest`s from the YAML. Noting this for future.

For now I’ve renamed the variable and added a comment in bc9103b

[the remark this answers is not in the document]
benthorner: Nice, thanks!

PR #207, opened 2021-12-30
Add PR CD warning and known testing caveat

benthorner:
We have decided this is an acceptable risk:

 - Most of the functionality is covered by the smoke tests and checking the Celery worker starts.

 - Adding functional tests to the pipeline would mean copy / pasting config from the -aws repo.

 - We wouldn't be able to prevent concurrent runs of the functional tests across pipelines [1].

 - The only functionality not covered is the worker running being able to run tasks from the queue.

[1]: https://github.com/concourse/concourse/issues/2691

### DIFF
--- a/.github/pull_request_template.yml
+++ b/.github/pull_request_template.yml
@@ -1,0 +1,8 @@
+
+
+
+---
+
+🚨⚠️ This will be deployed automatically all the way to production when you click merge ⚠️🚨
+
+See [docs/deploying.md](docs/deploying.md) for notes and caveats about this app.
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -1,0 +1,9 @@
+# Deploying
+
+See [the central documentation about merging and deploying](https://github.com/alphagov/notifications-manuals/wiki/Merging-and-deploying#deployment).
+
+## Caveats
+
+### The [deployment pipeline](https://github.com/alphagov/notifications-broadcasts-infra/blob/main/ci/govuk-alerts.yml) does't run functional tests for creating / approving alerts.
+
+You should manually test any changes that might affect the way the Celery worker runs or consumes from its queue e.g. changes to environment variables. This can be done locally, by deploying a branch or doing non-continuous (manual) deployment.


CrystalPea (approved):

docs/deploying.md
```diff
@@ -0,0 +1,9 @@
+# Deploying
+
+See [the central documentation about merging and deploying](https://github.com/alphagov/notifications-manuals/wiki/Merging-and-deploying#deployment).
+
+## Caveats
+
+### The [deployment pipeline](https://github.com/alphagov/notifications-broadcasts-infra/blob/main/ci/govuk-alerts.yml) does't run functional tests for creating / approving alerts.
+
+You should use a branch deployment to manually test any changes that might affect the way the Celery worker runs or consumes from its queue e.g. changing to environment variables.
```
idavidmcdonald, line 9: I don't think the lack of functional test coverage should mean the suggested action should be to do a branch deployment. I'd suggest that instead it would be to deploy it in our normal way but pause the staging pipeline as per the wiki so that you can run some manual testing before unpausing the staging pipeline when you are happy.

I could see an argument instead that if you can't test this locally with all the apps working together (can't remember if you can or can't) then you'd do the branch deployment.

.github/pull_request_template.yml
```diff
@@ -0,0 +1,8 @@
+
+
+
+---
+
+🚨⚠️ This will be deployed automatically all the way to production when you click merge ⚠️🚨
+
+See [docs/deploying.md](docs/deploying.md) for notes and caveats about this process.
```
idavidmcdonald, line 8: I know I wouldn't click on this link to read any additional docs as I would assume it is generic stuff from the wiki for a process I'm already familiar with that is shared across all our continuously deployed apps.

If there are caveats that people need to be aware of when merging a PR I suggest either we show it inline to them as part of the code they are changing or that it goes in this pull request template. Given how short the docs are below as well, moving it in to the pull request template wouldn't be too long I don't think.

idavidmcdonald (commented):
Suggesting a change of structure here please (although it is not a hill I'm going to die on if you feel super strongly about the current structure)

docs/deploying.md
benthorner: That's fair - I should have said "some kind of manual testing". I've reworded to explain. I don't think we should encourage "merge and hope", as that can unnecessarily block the master branch - it's quick enough to do a branch deployment, or a local test. I've suggested all three approaches.

(I'm pretty sure the local Celery approach works, but I can't test.)

.github/pull_request_template.yml
benthorner: > I know I wouldn't click on this link to read any additional docs

I would the first time, but not subsequent times. That's probably the most value we can expect to get out of this. I've changed "process" to "app", though. Would that be more of a trigger for you?

I'm not a fan of PR templates: they usually add noise. However, it does seem like the best mitigation for the risk here. I'd prefer to keep it as short as possible, by avoiding low-level guidance that could grow. I agree it's not _that_ long right now, but I can imagine it getting longer e.g. more pointers, or more caveats coming along. I don't want to have to scroll past a load of boilerplate every time I read someone's PR - [here's a past example I regret](https://github.com/alphagov/smokey/pull/695).

Since these are both subjective views, maybe we can just try it and see if someone goes "I would have done that testing if the warning had been more in-my-face 😄". What do you think?

CrystalPea: I actually agree with David here, I wouldn't read those docs either, but I would read a PR template. So it would be great to have something in the PR template, even if it's a pointer to the docs, like: "Caution, when you hit merge, this will deploy your changes all the way to prod, for more info about deployment process for this app see the README".

idavidmcdonald: As you say, it is subjective and not something I mind too much given the short and low cost nature of having this extra documentation.

I think I'd still suggest putting it all in the PR template but will approve this anyway because I don't think it is a high enough impact change to warrant too lengthy a discussion for something that is subjective and we have little evidence for either way.

idavidmcdonald (approved):
Nice one

docs/deploying.md
idavidmcdonald: Thanks, that is indeed an improvement

.github/pull_request_template.yml
benthorner: > for more info about deployment process for this app see the README

@CrystalPea I think the latest template should do just that 👌.

> I'd still suggest putting it all in the PR template but will approve this anyway

I get you. Given your comments I'd do that if I started again, but it's not worth it now.

PR #211, opened 2022-03-02
Update JS build environment to use Node 16 LTS & NPM 8.3.1

rjbaker:
Updates the version of Node required to build the project to the latest Node LTS version.
This upgrades the `package-lock.json` file to v2 syntax.

Additionally updates the Makefile bootstrap target to check if "n" (Node version manager) is installed before running
`n auto` to install the version specified in `package.json`.

In CI this will install the required Node version at build time instead of relying on the version bundled in the Concourse build image. See alphagov/notifications-aws/#1044

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ run-flask:
 .PHONY: bootstrap
 bootstrap:
 	pip install -r requirements_for_test.txt
+	$(if $(shell which n), n auto)
 	npm ci
 	npm run build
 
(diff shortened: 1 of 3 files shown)


benthorner (commented):
Confirmed everything still works locally as expected (JS & CSS) ✅.

Worth holding until [we're clear what we're doing with `n` though](https://github.com/alphagov/notifications-admin/pull/4178#pullrequestreview-897482019).

rjbaker:
Superseded by #217

PR #217, opened 2022-04-20
Build frontend/JS assets using Node 16 LTS

rjbaker:
This change makes the Node / NPM version and installation via NVM consistent with other front end projects.

 - Updates project to use Node 16 LTS
 - Updates `package-lock.json` to v2 format
 - Uses NVM to install Node version specified in `.nvmrc` file

See related PRs:

 - alphagov/notifications-admin#4202
 - alphagov/document-download-frontend#114


.nvmrc
```diff
@@ -0,0 +1 @@
+16.14.0
```
tombye, line 1: According to the [docs on nvm's repo](https://github.com/nvm-sh/nvm#nvmrc): 'The contents of a .nvmrc file must be the <version> (as described by nvm --help) followed by a newline.' and it looks like this doesn't have a newline.

tombye (approved):
Only spotted one, very minor, thing so approving this with the assumption that it'll be fixed before deploying.

.nvmrc
rjbaker: It doesn't appear to matter but i've fixed it anyway.

PR #221, opened 2022-05-06
Fix incorrect group for NPM dependencies

benthorner:
https://www.pivotaltracker.com/story/show/181889734

This means we can use tools like "npm audit" to look for security
vulnerabilities we definitely need to fix as they could pose a
direct risk to users. I've checked each of them with @tombye and
also against an external set of principles [^1].

[^1]: https://betterprogramming.pub/is-this-a-dependency-or-a-devdependency-678e04a55a5c

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ run-flask:
 .PHONY: bootstrap
 bootstrap:
 	pip install -r requirements_for_test.txt
-	source $(HOME)/.nvm/nvm.sh && nvm install && npm ci && npm run build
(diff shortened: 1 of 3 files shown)


tombye (approved):
Looks good to me 👍🏻 

I can see why I put `govuk-frontend` as a devDependency, because only the imported bits are shipped and I was thinking of which packages are sent as-is. I do think the opposite argument is stronger with hindsight, as the code from it is still shipped as-is, just not all of it, so this seems correct.

PR #223, opened 2022-05-23
Make URLs in alerts clickable 

quis:
Should not be merged as-is, should reuse code from `notifications-utils`

---


tests/app/main/views/test_alert.py
```diff
@@ -73,3 +73,34 @@ def test_alert_says_active_alert_is_active(client_get, mocker):
...
+    (
+        'check-for-flooding.service.gov.uk/target-area/055FWFPUNE07',
+        'http://check-for-flooding.service.gov.uk/target-area/055FWFPUNE07',
+    ),
+    (
+        'https://check-for-flooding.service.gov.uk/target-area/055FWFPUNE07',
+        'https://check-for-flooding.service.gov.uk/target-area/055FWFPUNE07',
+    ),
+))
+def test_urls_in_alerts_are_clickable(client_get, mocker, url, expected_href_attribute):
```
benthorner, line 94: Could we make do with just one test that the happy path for autolinking works? [It looks like we're testing all the edge cases in utils](https://github.com/alphagov/notifications-utils/pull/971/files#diff-f28b3ed23223b000771c34964f74ea35379b271cc8b4fe4191cd0c637b621fbd), so I'm not sure what value there is in testing a selection of them again.

Do we need them all?

quis: Yup, fair point, I’ve cut it down to one simple case and one complex case.

benthorner (approved):
Confirmed this works locally on all pages with one of the alerts in data.yaml ✅.

PR #245, opened 2023-02-17
Fix planned test status box out of sync with planned tests page

Adam-C-Mitchell:
The following fixes the issue where a planned test will be shown in the status box, but not reflected on the planned tests page when clicked through to it. This is due to the status box checking only for a match in the date (thereby allowing the planned test to display all day), while the planned tests page checks the time for expiry. The status box now uses the planned function to verify whether the test alert has expired.

---

### DIFF
--- a/app/models/alerts.py
+++ b/app/models/alerts.py
@@ -90,16 +90,12 @@ class Alerts(SerialisedModelCollection):
         }
 
     @property
-    def planned_non_public_tests(self):
-        return [
-            planned_test for planned_test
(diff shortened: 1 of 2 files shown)


gds-gharper (approved):
So having checked I'm good with the logic for now (We will want to change it later).  However I feel a change like this should break a test/require test updates.

Can we get a test case for this?

Adam-C-Mitchell:
I'll raise a PR with a few test cases for this shortly

PR #247, opened 2023-02-27
EAS 713 - Service Tests Flow / Announcements

Adam-C-Mitchell:
This changes the flow of service tests through the application, by:

- Displaying the service test in the status box on the home page for 1 hour, superseding an announcement (in the planned-tests.yaml) but still superseded by a current alert on a public channel.
- Displaying the service test(s) on the service tests page - which replaces the planned tests page and should be redirected as such - until the end of that day, after which point they're moved into past alerts.
- Holding service tests in past alerts for 48 hours before they are considered 'expired' and no longer displayed. This does not apply for alerts sent on public channels, which still remain in past alerts indefinitely.
- Displaying announcements - i.e., planned tests from the planned-tests.yaml in the announcements page for as long as they haven't expired (i.e., finished or cancelled). The announcements page **DOES NOT** display service tests, and the service tests page **DOES NOT** display announcements. Only the service tests page is linked to from related content - only an active planned test status box will link to the announcements page.

---


app/templates/views/announcements.cy.html
```diff
@@ -0,0 +1,129 @@
+{% extends "content.html" %}
+
+{%- from "govuk_frontend_jinja/components/breadcrumbs/macro.html" import govukBreadcrumbs -%}
+{%- from "govuk_frontend_jinja_overrides/templates/components/language-select/macro.html" import hmrcLanguageSelect -%}
+{%- from "components/related_content.html" import related_content -%}
+{%- from "components/meta_tags.html" import metaTags -%}
+
+{% set pageTitle = "Profion ar y gweill" %}
```
gds-gharper, line 8: This looks to be the title from the planned tests page, it'll need translating properly, along with anything else new on this page.

gds-gharper (commented):
Liking the new tests

app/templates/views/announcements.cy.html
Adam-C-Mitchell: Ah, good shout. Okay, let's leave this be for now if there are no other issues, and I'll see if I can get those back shortly and added to the PR.

gds-gharper (approved):

PR #251, opened 2023-03-06
EAS-447 Add docker file and build script for gov.uk/alerts

jonathan-owens-gds:
In preparation for migration to the Github CI pipeline and developer work-flow containerization, a Dockerfile has been added to this repo, together with a build script to allow multi-platform builds to be produced, if necessary, and then pushed to ECR or imported to the local docker host.

For the container to serve the gov.uk/alerts website on port 6017, the container will need to be launched within a network giving the application access to the api on port 6011. Note: This will be developed in future work.

---


simon-sorrell-gds-alerts (commented):
Looks fine to me, but deferring to Lewis to approve just to be certain that the python code updates will have no potential impact to the current live processes

scripts/aws-creds.sh
```diff
@@ -0,0 +1,19 @@
+#! /bin/bash
```
lstevens-cabinet-office, line 1: cant see why this is required as i cannot see any files reference it.

scripts/docker-build.sh
```diff
@@ -0,0 +1,36 @@
+#! /bin/sh
```
gds-gharper, line 1: TIL you can have spaces between the shebang and the command

scripts/docker-build.sh
```diff
@@ -0,0 +1,36 @@
...
+while [ $# -gt 0 ]; do
+  if [[ $1 == *"--"* ]]; then
+    param="${1/--/}"
+    declare $param="$2"
+  fi
+  shift
+done
+
+if [[ $ENVIRONMENT == "development" ]]; then
+  ECS_ACCOUNT_NUMBER=388086622185
```
gds-gharper, line 16: Should we be hard coding the account numbers in this repo like this rather than fetching them from somewhere else?

scripts/aws-creds.sh
jonathan-owens-gds: file removed

PR #263, opened 2023-04-04
EAS-862 Add National Test time to banner and announcement

jonathan-owens-gds:
To add the time to the landing page banner and announcement page, the following changes were made:

- Add a method to convert the meridiem indicator to lowercase
- Use new alert_date method to display date & time in banner and announcement templates
- Update alert-date and announcements test files to reflect the extra 'time' text
- Update planned test times in the dev, preview, staging and prod YAML files (in UTC)

---

### DIFF
--- a/app/models/alert_date.py
+++ b/app/models/alert_date.py
@@ -69,11 +69,6 @@ class AlertDate(object):
         return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y} at {dt:%-I%p}'
 
     @property
-    def as_alert_date_lang(self, lang='en-GB'):
-        dt = self._local_datetime
-        return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y}'
-
-    @property
     def as_iso8601(self):
         return self._local_datetime.isoformat()
 
--- a/app/models/alert_date.py
+++ b/app/models/alert_date.py
@@ -42,6 +42,12 @@ class AlertDate(object):
         return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y}'
 
     @property
+    def datetime_as_lang(self):
+        dt = self._local_datetime
+        meridiem = f'{dt:%p}'.lower()
+        return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y} at {dt:%-I}{meridiem}'
+
+    @property
     def as_url(self):
         """
         * non-zero padded day
@@ -60,7 +66,7 @@ class AlertDate(object):
     @property
     def as_alert_lang(self, lang='en-GB'):
         dt = self._local_datetime
-        return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y}, at {dt:%-I%p}'
+        return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y} at {dt:%-I%p}'
 
     @property
     def as_alert_date_lang(self, lang='en-GB'):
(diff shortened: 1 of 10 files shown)


app/templates/components/banner.html
```diff
@@ -18,7 +18,7 @@ <h2 class="alerts-notification-banner__title govuk-heading-m govuk-!-margin-bott
   {% set title %}
     {% for alert_date in dates_of_test_alerts|sort %}
       {% if loop.index == 1 %}
-        On {{ alert_date.as_alert_date_lang }}, there will be a national test of the UK Emergency Alerts service.
```
gds-gharper, line 21: I added this function `as_alert_date_lang` and it is only used for this place in the code and nowhere else.  As such you can just replace the format inside that function (or just delete the function).

app/templates/views/announcements.cy.html
```diff
@@ -100,7 +100,7 @@ <h1 class="govuk-heading-xl">
     {% endif %}
     {% if loop.first %}
       <h2 class="govuk-heading-m {% if not date_loop.first %}govuk-!-margin-top-9{% endif %}">
-        {{ alert_or_planned_test.starts_at_date.date_as_lang }}
+        {{ alert_or_planned_test.starts_at_date.datetime_as_lang }}
```
gds-gharper, line 103: I think macro loop around this breaks down the planned tests by date.  As such there's no guarantee the second planned test will be at the same time as the first planned test

app/models/alert_date.py
```diff
@@ -60,7 +66,7 @@ def as_lang(self, lang='en-GB'):
     @property
     def as_alert_lang(self, lang='en-GB'):
         dt = self._local_datetime
-        return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y}, at {dt:%-I%p}'
+        return f'{dt:%A} {dt.day} {dt:%B} {dt:%Y} at {dt:%-I%p}'
```
gds-gharper, line 69: Is the only difference between this and the function above that one has AM in capitals and the other has AM in lowercase?

If that's the case I think we may as well just settle on one format and use it for all of them.

gds-gharper (changes requested):

app/models/alert_date.py
jonathan-owens-gds: This particular line just removes the comma - in the soft launch on the 19th Mar, it was suggested that the announcement banner had too many commas and didn't scan well.

Regarding the upper/lower case methods: As the uppercase has been the only option until now, I assume that this is in line with the GOV.UK style guide.  This looked a little jarring in the banner and announcement headers, which is why I implemented a lowercase option. I think we should still keep both options.

gds-gharper: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates

I don't think we should have 2 different options, we should have one and stick with it.  We can check with the content team to decide which one.

app/templates/views/announcements.cy.html
jonathan-owens-gds: The page is only designed to show one planned alert, with the sections allowing the announcement to be different for the different countries of the UK.  If the page needs to be modified to allow multiple planned alerts on different dates to be shown, this will need to go through UI prototyping first.

app/templates/components/banner.html
jonathan-owens-gds: Unused method removed

app/models/alert_date.py
jonathan-owens-gds: Mehmet confirmed that "gov.uk content" team were happy with the lowercase am/pm

gds-gharper (approved):

PR #283, opened 2023-05-15
Updating the dockerfile for govukalerts

lstevens-cabinet-office:
---


simon-sorrell-gds-alerts (approved):
Updates to the dockerfile are fine, as they should fix the bits that were broken

PR #293, opened 2023-05-17
EAS-1070 Upload assets with mime_type metadata

jonathan-owens-gds:
Assets should be uploaded to S3 with mime type metadata so the server knows what to do with them.

---


simon-sorrell-gds-alerts (approved):
Let's give it a go as it isn't working right as it is anyway, so fine to try it

PR #308, opened 2023-06-29
Update system testing page content and display service tests

Adam-C-Mitchell:
---

### DIFF
--- a/app/models/alert.py
+++ b/app/models/alert.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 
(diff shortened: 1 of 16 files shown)


gds-gharper (approved):
Do we need to update the smoke tests at all?

Adam-C-Mitchell:
> Do we need to update the smoke tests at all?
We won't, but we will need to change the redirects, so I will raise a separate PR shortly to remove the planned tests page since we haven't been using that for a while, change the redirects, then remove the service tests page once those have taken effect

PR #330, opened 2023-11-06
EAS-1576 Add config for local running

jonathan-owens-gds:
The HOST variable is used to distinguish between running locally and on the hosted infrastructure (i.e. AWS). This variable can therefore take one of the following values:

HOST = [ local | hosted | test ]

"local" indicates that the service will be configured for running on a local machine. "hosted" is intended for use when the service is running on the AWS-hosted infrastructure. "test" provides a special set of configuration values that are used by the unit, integration and functional tests.

The environment variable ENVIRONMENT is used to tell the service which set of config values to take up, and can be set to one of the following values:

ENVIRONMENT = [ local | development | preview | staging | production ]

A value of "local" indicates that the service will be running on the development machine. A value corresponding to any of the others in the above set maps directly to the name of the environment hosted in AWS.

The development environment hosted on AWS will now configure the above variables as follows:
HOST=hosted & ENVIRONMENT=development

---

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -20,7 +20,7 @@ def create_app():
     )
 
     from app.config import configs
-    environment = os.getenv('HOST', 'development')
+    environment = os.getenv('HOST', 'local')
     application.config.from_object(configs[environment])
 
     from app.main import bp as main
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -20,7 +20,7 @@ def create_app():
     )
 
     from app.config import configs
-    environment = os.getenv('NOTIFY_ENVIRONMENT', 'development')
+    environment = os.getenv('HOST', 'development')
     application.config.from_object(configs[environment])
 
     from app.main import bp as main
--- a/app/config.py
+++ b/app/config.py
@@ -110,16 +110,16 @@ class Development(Config):
     PLANNED_TESTS_YAML_FILE_NAME = "planned-tests-dev.yaml"
 
 
-# class Test(Config):
-#     DEBUG = True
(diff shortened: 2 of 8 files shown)


app/__init__.py
```diff
@@ -20,7 +20,7 @@ def create_app():
     )
 
     from app.config import configs
-    environment = os.getenv('NOTIFY_ENVIRONMENT', 'development')
+    environment = os.getenv('HOST', 'development')
```
Adam-C-Mitchell, line 23: Shouldn't this default to "local", since development is now not an option for the host?

app/config.py
```diff
@@ -103,13 +103,6 @@ class Decoupled(Config):
     PLANNED_TESTS_YAML_FILE_NAME = "planned-tests.yaml"
```
Adam-C-Mitchell, line 103: With regards to these files I think we should either be removing the environmentally-conditional ones (i.e., "planned-tests-<dev|preview|staging>.yaml" or changing this to something like PLANNED_TESTS_YAML_FILE_NAME = `f"planned-tests-{os.getenv('ENVIRONMENT')}.yaml"`
It was previously desired to see what planned tests would look like across environments, but this may have been at a time where we didn't have manual approval of production code, thus they may be redundant files now.

Adam-C-Mitchell (changes requested):

Adam-C-Mitchell:
Otherwise this looks good to me, some good changes that definitely make it easier to understand configs across the infrastructure I think.

Adam-C-Mitchell (approved):

PR #346, opened 2024-03-05
EAS-2000 Update pyyaml version

jonathan-owens-gds:
- Update pyyaml version (to avoid conflicts when building api/admin/govuk)
- When building locally, the local version of Utils is used, instead of pulling from Github (this follows the way all the other repos do this now)


Adam-C-Mitchell (approved):
Looks good along with the corresponding changes in the other repos provided they're looking good in a hosted env as well

PR #349, opened 2024-03-27
Implementation of capitalize util and usage for alerts

Adam-C-Mitchell:
Simple implementation of a capitalisation util to ensure any alerts displayed have the first character of their display name capitalised (sometimes they will not be when retrieved from the database, for instance when using custom polygons).


leylayaltiligil (approved):
Tests passing.

PR #354, opened 2024-05-08
EAS-2030 Simplification of Custom Area names displayed

leylayaltiligil:
The removal of redundant detail in Custom Area names when displayed on gov.uk/alerts and the relevant tests for this. 
For example, "5km around the easting of 530111.0 and the northing of 171161.0, in Lambeth" is displayed as "an area in Lambeth".

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -16,13 +16,15 @@ def capitalise(value):
 
 
 def simplify_custom_area_name(value):
+    # Checking for words indicating that it is a
+    # custom area OR that is has a local authority in the name
     if (
         'postcode' not in value
         and 'easting' not in value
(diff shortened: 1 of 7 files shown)


app/utils.py
```diff
@@ -15,6 +15,31 @@ def capitalise(value):
...
+
+
+def is_custom_area_with_local_authority(value):
+    return (
+        (
+            'postcode' in value
+            or 'easting' in value
+            or 'latitude' in value
+        ) and (" in " in value)
+    )
```
gds-gharper, line 40: This won't work for welsh, we should be able to get it translated easily enough by reaching out

gds-gharper (approved):

PR #365, opened 2024-10-07
Abstract regions, account numbers and tags from buildspec

Adam-C-Mitchell:


Dockerfile.eas-govuk-alerts
```diff
@@ -1,7 +1,8 @@
 ARG ECS_ACCOUNT_NUMBER
 ARG RESOURCE_PREFIX=eas-app
+ARG AWS_REGION=eu-west-2
```
simon-sorrell-gds-alerts, line 3: Just wanted to make sure I understand this. I noted ECS_ACCOUNT_NUMBER doesn't look like it is being explicitly set in the Dockerfile, so I am guessing it is being passed into the build as an environment variable? If it is, couldn't/shouldn't the region also be passed in rather than being set in the Dockerfile itself? I have a feeling I am missing something, so please feel free to tell me I am talking rubbish :-)

Adam-C-Mitchell: All of those args will be passed in explicitly anyway, the assignments you see (e.g., AWS_REGION=eu-west-2) is just a default - so the only one that technically needs to be passed in is ECS_ACCOUNT_NUMBER, which we don't default on because that is always going to be different per environment (whereas the region, resource prefix etc., are generally the same everywhere). Either way, as far as the buildspec is concerned they'll always be explicitly passed in from environment variables.

simon-sorrell-gds-alerts (dismissed):

simon-sorrell-gds-alerts (approved):

PR #371, opened 2025-01-01
Correct a spelling mistake

EdwardBetts:

PR #372, opened 2025-01-01
Avoid utcnow() which is deprecated

EdwardBetts:
Replaced the use of the deprecated datetime.utcnow() method with
datetime.now(datetime.timezone.utc) to create an aware datetime object
representing the current UTC time.


jonathan-owens-gds (changes requested):
Hi Edward, thanks for your contribution. If you could sign your commit, as per #371, I'd be happy to merge. Thanks

PR #377, opened 2025-02-03
EAS-1055 Log broadcast-event-id with each page upload

jonathan-owens-gds:
As per the docs: "The number of worker processes/threads can be changed using the --concurrency argument and defaults to the number of CPUs available on the machine."

The `--autoscale` argument keeps a minimum of 1 process, but permits scaling up to 8 processes to handle queued tasks.


simon-sorrell-gds-alerts (approved):
Any additional visibility/information added to logging is welcomed

Adam-C-Mitchell (approved):

PR #379, opened 2025-02-13
EAS-2552 Feed stylesheet, incl. Welsh

jonathan-owens-gds:
The Atom feed now has styling that is applied when viewing the file in a browser. The page also has a Welsh stylesheet that is linked from the Welsh pages on the gov.uk/alerts site


app/render.py
```diff
@@ -182,7 +229,29 @@ def _add_feed_entry(fg, alert, alert_url):
...
     fe.author(name="Emergency Alerts Service", uri="https://www.gov.uk/contact/govuk")
     fe.content(alert.content)
-    fe.link(href=f"{host_url}/alerts/" + alert_url)
+    fe.link(href=f"{host_url}/alerts/" + alert_url, rel="alternate")
     content = alert.content if len(alert.content) <= 40 else alert.content[:36] + "..."
     fe.summary(content)
     fe.published(alert.approved_at)
+
+
+def _add_stylesheet_attribute_to_atom(
```
gds-gharper, line 238: This feels like parsing XML with regular expressions.  It feels like the sensible way of doing this is with a custom extension to FeedGenerator https://feedgen.kiesow.be/#extensions

This is what creates the atom https://github.com/lkiesow/python-feedgen/blob/main/feedgen/feed.py#L86

So it looks like you to subclass https://github.com/lkiesow/python-feedgen/blob/main/feedgen/ext/base.py and then have extend_atom insert the xml-stylesheet entry. https://docs.python.org/3/library/xml.etree.elementtree.html#element-objects

It looks like it's not quite as trivial as that, but there's a stack overflow here which looks like it'll do the job
https://stackoverflow.com/questions/26442708/how-to-keep-the-xml-stylesheet.  The only difference is you'll be getting a root element.

You might then be able to write the stylesheet tag by using addprevious as hinted at here https://post.bytes.com/forum/topic/python/746870-lxml-and-adding-a-stylesheet

However all of this is a little guesswork, the easiest way of accomplishing might be  to write the xml declaration yourself.

https://github.com/lkiesow/python-feedgen/blob/main/feedgen/feed.py#L204

If you pass in xml_declaration=False to this function, then you will get the raw XML without the xml declaration.  Then you can simply write both the xml_tag and the stylesheet_tag followed by the feed_string and you will have a correctly formed xml file with the stylesheet tag.

gds-gharper (approved):

PR #393, opened 2025-06-17
EAS-2629 Python upgrade

jonathan-owens-gds:
Upgrade Python to version 3.12.11 (the actual upgrade comes in via the base image - package upgrades applied to this repo are to maintain compatibility).


jamesorlakin (dismissed):

.github/workflows/on-pr-into-main.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
       - name: Install and upgrade pre-requisites
         run: |
-          python3.9 -m pip install --upgrade pip wheel setuptools
+          python3.12 -m pip install --upgrade pip wheel setuptools
```
gds-gharper, line 30: Would it be possible to make it so we don't need to update this file in 2 places to keep the versions in sync?

Makefile
```diff
@@ -124,16 +127,21 @@ test:
...
-		--loglevel=WARNING \
-		--concurrency=1 \
-		--autoscale=1,1
+		--loglevel=INFO \
+		--autoscale=8,1 \
+		--hostname='govuk-alerts@%h'
+
+.PHONY: uninstall-packages
+uninstall-packages:
+	python -m pip uninstall emergency-alerts-utils -y
```
gds-gharper, line 146: Is there a reason we need the uninistall command?

app/celery/tasks.py
```diff
@@ -13,7 +14,9 @@ def publish_govuk_alerts(self, broadcast_event_id=""):
     try:
         alerts = Alerts.load()
         rendered_pages = get_rendered_pages(alerts)
-
+        if os.environ.get("HOST") == "local":
+            current_app.logger.info("Skipping upload to S3 in local environment")
+            return
```
gds-gharper, line 19: Would it make more sense to make this a config option in `config.py` pulled from the environment rather than directly checking the environment here?  Something like `DISABLE_S3_UPLOAD`

That said, it might make more sense to move the `GOVUK_ALERTS_S3_BUCKET_NAME` config option only into the Hosted config and check for the existence of that, if it's not found don't upload to S3.

app/utils.py
```diff
@@ -81,14 +81,13 @@ def is_in_uk(simple_polygons):
 def upload_html_to_s3(rendered_pages, broadcast_event_id=""):
```
gds-gharper, line 81: Perhaps make the same check here, check config to see if `GOVUK_ALERTS_S3_BUCKET_NAME` is set and if it's not sent make this function return early.

gds-gharper (changes requested):

.github/workflows/on-pr-into-main.yml
jonathan-owens-gds: command now uses ${{ env.PYTHON_VERSION }}

Makefile
jonathan-owens-gds: This is very useful locally - I've found that when "freezing" the package requirements, pip-compile will try to reconcile package upgrades with the currently installed packages. This can lead to variances between local installations and those installed on a clean instance (in github actions, AWS containers etc.). This "uninstall packages" command is intended to be run locally to ensure pip package resolution gives an import tree that is the same as that on hosted machines.

app/utils.py
jonathan-owens-gds: check here also for presence of GOVUK_ALERTS_S3_BUCKET_NAME env var

jamesorlakin (approved):

app/utils.py
```diff
@@ -117,7 +117,8 @@ def upload_assets_to_s3():
     if not Path(DIST).exists():
         raise FileExistsError(f'Folder {DIST} not found.')
 
-    bucket_name = os.environ.get("GOVUK_ALERTS_S3_BUCKET_NAME")
+    # bucket_name = os.environ.get("GOVUK_ALERTS_S3_BUCKET_NAME")
```
jamesorlakin, line 120:
```suggestion
```

jamesorlakin (approved):

app/utils.py
jonathan-owens-gds: comment removed

app/config.py
```diff
@@ -24,45 +23,41 @@ class Config():
...
-        "worker_max_tasks_per_child": 20
+        "task_queues": [Queue(QUEUE_NAME, Exchange("default"), routing_key=QUEUE_NAME)],
+        "worker_max_tasks_per_child": 2,
     }
 
     PLANNED_TESTS_YAML_FILE_NAME = "planned-tests.yaml"
 
 
 class Hosted(Config):
+    HOST = os.environ.get('HOST')
```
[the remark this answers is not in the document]
jonathan-owens-gds: agreed - code removed

gds-gharper (approved):

PR #423, opened 2025-09-03
GOVUK Prepare content changes

leylayaltiligil:
Removal of gov.uk/prepare link and link replaced, addition of image (TBD)


app/assets/images/prepare-banner.png
gds-gharper, line 1: I like we've optimised it down from the nearly 1MB on the prepare site

gds-gharper (approved):

PR #517, opened 2026-04-10
EAS-2889: Adjust CAP XML generation for feed (includes cancellation)

leylayaltiligil:
This PR consists of the following:
- Adjustment to the CAP cancel alert generation functionality - resulting in CAP XML with `references` element to link back to previous alert
- Addition of functionality populating the `areaDesc` element with the Alert's`display_areas` text
- Adjustment to `area` element generation - now one area may have multiple polygons so the `areaDesc` is relevant for entire area
- Removes local from `language` element - as instructed during Alert Hub call
- CAP XML files published to different directory from other files

### DIFF
--- a/app/render.py
+++ b/app/render.py
@@ -355,13 +355,12 @@ def get_cap_xml_for_alerts(alerts, publish_task_progress=None):
 
         # If alert has been cancelled, generate another CAP XML file for the updated event
         if alert.cancelled_at:
-            new_identifier = str(uuid.uuid4())
             event = create_cap_event(
                 alert,
-                new_identifier,
+                f"{identifier}-1",
                 url=alert_url_with_host,
                 cancelled=True,
-                prev_alert_identifier=identifier
+                prev_alert_identifier=identifier,
             )
             cap_xml = generate_xml_body(event)
             timestamp = alert.cancelled_at.strftime("%Y%m%d%H%M%S")
--- a/app/render.py
+++ b/app/render.py
@@ -351,14 +351,14 @@ def get_cap_xml_for_alerts(alerts, publish_task_progress=None):
         event = create_cap_event(alert, identifier, url=alert_url_with_host)
         cap_xml = generate_xml_body(event)
         timestamp = alert.approved_at.strftime("%Y%m%d%H%M%S")
-        cap_xml_alerts[f"alerts/{alert_url}-{timestamp}.cap.xml"] = cap_xml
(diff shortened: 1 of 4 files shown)


Adam-C-Mitchell (dismissed):

Adam-C-Mitchell (dismissed):

app/render.py
```diff
@@ -351,14 +351,21 @@ def get_cap_xml_for_alerts(alerts, publish_task_progress=None):
         event = create_cap_event(alert, identifier, url=alert_url_with_host)
         cap_xml = generate_xml_body(event)
         timestamp = alert.approved_at.strftime("%Y%m%d%H%M%S")
-        cap_xml_alerts[f"alerts/{alert_url}-{timestamp}.cap.xml"] = cap_xml
+        cap_xml_alerts[f"cap-xml/{alert_url}-{timestamp}.cap.xml"] = cap_xml
 
         # If alert has been cancelled, generate another CAP XML file for the updated event
         if alert.cancelled_at:
-            event = create_cap_event(alert, identifier, url=alert_url_with_host, cancelled=True)
+            new_identifier = str(uuid.uuid4())
```
jamesorlakin: I think we'd want this identifier to be stable so it's not seen as a 'new' update to a previous reference upon every rebuild?

E.g. for cancellations we could have the previous identifier plus `-1` or something?

jamesorlakin (dismissed):

app/render.py
gds-gharper: Yes, all CAP XML generations should be repeatable and regeneratable.  The filename should never change regardless of the state of the s3 bucket.

gds-gharper (changes requested):

app/render.py
leylayaltiligil: @jamesorlakin  The identifiers are to be changed once we have a registered OID (maybe we use this and the event ID?).
I kept this here as a placeholder and just because identifiers need to be unique per alert but can just append '-1' on the end of whatever previous alert identifier is.
@gds-gharper  The identifier isn't part of the filename, it won't change as it's a string made up of **{`alert_url`}_{`approved_at` or `created_at` timestamp}.cap.xml,** where `alert_url` is the alert url slug e.g. `24-apr-2026-2` (same as html alert slug).

gds-gharper: Ah, my mistake

gds-gharper (approved):

PR #522, opened 2026-04-24
EAS-2937: create website archive file following successful publish

Dave-Johnson-GDS:


app/commands.py
```diff
@@ -23,10 +24,11 @@ def setup_commands(app):
 def publish():
     try:
         publish_task_progress = PublishTaskProgress.create(publish_type="publish-dynamic", publish_origin="cli")
-        _publish_html(publish_task_progress)
+        published_html = _publish_html(publish_task_progress)
```
gds-gharper, line 27: I don't think this is an issue you've created, but when we do a partial publish do we not publish the CAP XML as well?  We certainly should be, and it needs to be included in the archive.

app/commands.py
```diff
@@ -23,10 +24,11 @@ def setup_commands(app):
...
         purge_fastly_cache()
         alerts_api_client.send_publish_acknowledgement()
         publish_task_progress.set_to_finished()
+        archive_website(html=published_html)
```
gds-gharper: Hmm, in this case, when we're only doing a partial publish (no assets) we need to handle this differently, otherwise the archive won't contain a full site.

I'm not totally sure how we should handle this.  I'm torn between grabbing the previous archive and using the files from there (probably faster, but would break if there's no previous archive) and grabbing the assets from S3 one at a time.  Interested ot hear other opinions or options.

gds-gharper (changes requested):

app/commands.py
Dave-Johnson-GDS: There is a bit of a discrepancy here - this partial publish is only ever initiated by one of us hopping on to a container and running the flask publish command, the celery task publishes both html and CAP XML.  I think this should be updated to also include CAP XML.

app/commands.py
Dave-Johnson-GDS: The archive_website() function will retrieve assets if none have been passed in to it, so you should get a full publish once CAP XML is included.

gds-gharper (dismissed):

Adam-C-Mitchell (approved):

PR #523, opened 2026-04-29
EAS-3021: Sources `sender` value for CAP XML from environment variable or provided

leylayaltiligil:
This PR consists of changes regarding the CAP XML generation, specifically setting the `sender` element value to either what has been provided (will possibly be provided in the future), else use the default which is sourced from config.


gds-gharper (commented):
We probably want to draft this since it's not obvious we should be using an email, it might be a web address like the met office do

leylayaltiligil:
Closing this as the work, requiring this, has been paused.

PR #536, opened 2026-06-22
EAS-2957: govuk publish blue green deployment

Dave-Johnson-GDS:
`utils:EAS-2957-blue-green-web-deployments`
`api:EAS-2957-publish-blue-green-deployment`
`functional-tests:EAS-2957-blue-green-publish`


app/tasks/tasks.py
```diff
@@ -85,6 +121,90 @@ def publish_govuk_alerts(broadcast_event_id=""):
...
+        current_app.logger.info(f"Preparing publish destination {publish_destination}")
+        with tracer.start_as_current_span("Prepare publish destination"):
+            try:
+                # delete content from destination
+                prepare_destination(publish_destination)
+            except Exception as e:
+                current_app.logger.exception(f"Problem preparing publish site, will overwrite existing content: {e}")
+
+        # Set old cut_off date, so that all alerts are re-rendered
+        cut_off = datetime(1970, 1, 1, tzinfo=timezone.utc)
```
gds-gharper: I'm not sure I'd want to set the original epoch as a magic number in this case, I'd rather send through a null value and just disable the check

app/tasks/tasks.py
```diff
@@ -37,16 +43,43 @@ def publish_govuk_alerts(broadcast_event_id=""):
...
+                restore_latest_archive(publish_destination)
+            except Exception as e:
+                current_app.logger.exception(f"Problem preparing publish site, setting flag to run full publish: {e}")
+                prepared_ok = False
+
+        # Not able to confirm whether our publish destination was prepared correctly,
+        # therefore need to render and publish everything.
+        # Grab assets, and set a really old cut_off date.
+        if not prepared_ok:
+            cut_off = datetime(1970, 1, 1, tzinfo=timezone.utc)
```
gds-gharper: I think I'd rather not use a magic epoch value in this case, and instead send through an empty value and actively disable the check/where clause

gds-gharper (changes requested):

app/tasks/tasks.py
Dave-Johnson-GDS: Thanks - now set to None, which is already handled correctly by Jonathans code.

app/tasks/tasks.py
Dave-Johnson-GDS: Thanks - now set to None, which is already handled correctly by Jonathans code.

gds-gharper (dismissed):

Dave-Johnson-GDS:
Sorry @gds-gharper, had to tweak a unit test following the cut_off change.

app/tasks/tasks.py
```diff
@@ -85,6 +120,87 @@ def publish_govuk_alerts(broadcast_event_id=""):
...
+    setattr(g, FLASK_G_BROADCAST_EVENT_ID, broadcast_event_id)
+
+    try:
+        current_app.logger.info(
+            "Starting GovUK full publish. (Triggered by broadcast event: %s)",
+            broadcast_event_id,
+        )
+
+        publish_task_progress = PublishTaskProgress.create(
+            publish_type="publish-dynamic", publish_origin="dramatiq"
```
jamesorlakin: Should we differentiate this between full publish and partial?

app/utils.py
```diff
@@ -340,3 +351,238 @@ def archive_website(html, capxml, assets=None):
...
+            Body=file["data"],
+            ContentType=file["content_type"],
+        )
+        return file["key"]
+
+    with ThreadPoolExecutor(max_workers=max_workers) as pool:
+        futures = {pool.submit(upload_one, f): f["key"] for f in files}
+
+        for future in as_completed(futures):
+            future.result()  # propagate exceptions
```
jamesorlakin, line 588: Yay!

app/tasks/tasks.py
```diff
@@ -85,6 +120,87 @@ def publish_govuk_alerts(broadcast_event_id=""):
         raise e
 
 
+@dramatiq_instance.actor(
+    actor_name=TaskNames.PUBLISH_GOVUK_ALERTS_FULL,
+    queue_name=QueueNames.GOVUK_ALERTS,
+    allow_retry=True,
+)
+def publish_govuk_alerts_full(broadcast_event_id=""):
```
jamesorlakin, line 128: I almost feel this method is the same as the above, i.e. `cut_off=None` and not calling `restore_latest_archive`?

But I'm not worried enough that you'd need to rework it or anything.

jamesorlakin (dismissed):

app/tasks/tasks.py
Dave-Johnson-GDS: Good spot, now sorted.

app/utils.py
Dave-Johnson-GDS: :)

jamesorlakin (approved):

PR #541, opened 2026-07-08
EAS-3168 Fix date rendering in atom feed

jonathan-owens-gds:
The incorrect rendering of the "stopped at" date/time affected browsers only. The content in the atom feed has always been correct.

Legacy Javascript was parsing a datetime string from fields in the atom element. The "stopped at" date/time is already built out server-side, so we don't need javascript to construct it.
Tested in Chrome, Firefox and Safari.


gulpfile.js
```diff
@@ -135,7 +130,7 @@ const defaultTask = parallel(
   copy.govuk_frontend.images,
   copy.html5shiv,
   copy.images,
-  copy.plainjs,
+  // copy.plainjs,
```
jamesorlakin:
```suggestion
```

jamesorlakin (dismissed):

gulpfile.js
jonathan-owens-gds: done

jamesorlakin (approved):

